PR #34, opened 2015-11-27
Ejh doc

edhartnett:
Document examplePio.c. Also fix a bug in example code (decomposition array was not one-based). Also add a -v command line option to help debugging.

### DIFF
--- a/doc/Doxyfile.in
+++ b/doc/Doxyfile.in
@@ -745,6 +745,7 @@ WARN_LOGFILE           =
 
 INPUT                  = @CMAKE_CURRENT_SOURCE_DIR@/source \
                          @CMAKE_CURRENT_SOURCE_DIR@/../src \
+                         @CMAKE_CURRENT_SOURCE_DIR@/../examples/c \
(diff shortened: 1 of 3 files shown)


jedwards4b:
Hi Ed,

It just occured to me to tell you to look at the performance test as another
good, albeit more complicated, example. 

Looking through the pull request something doesn't seem quite right to me.

At line 221 the DataBuffer is assigned 
this->myrank + VAL
So I would expect the results for four procs to be:
{42,42,42,42,43,43,43,43,44,44,44,44,45,45,45,45}
and different from the result using 1,2,8 or 16 procs.

That's not consistent with line 550.  Also line 550 shows the final value as missing?

The decomp array is indeed 1 based, and 0 has a special meaning - it means to skip this array location 
in the read or write operation.

edhartnett:
Good catch! The documentation was incorrect in this case, it showed the ncdump from a file before I fixed the bug with the compdof array. I've checked in a fix that updates the docs and shows the ncdump for a 4-processor run, which is as you expect:
      foo = 42, 42, 42, 42, 43, 43, 43, 43, 44, 44, 44, 44, 45, 45, 45, 45 ;

PR #37, opened 2015-12-02
Ejh doc2

edhartnett:
Fortran .in Files
--------------------
In documenting examplePio_f I found that the fortran .in files were not being properly parsed by doxygen. I needed the documentation in those files, so I made some changes to get it into the final output.

Instead of looking at the .in files (whcih doxygen seems to have a hard time parsing) I changed it to look at the generated F90 files which are the result of the .in files being processed. This also allows the documentation for these functions to correctly reflect the type and number of dimensions in each case.

Documentation/Release Version Numbers
-----------------------------------------------------
I also added version numbers to the main CMakeList file, so that the documentation has the correct version number. (I guessed that the current ParallelIO release was 2.3.8 - not sure where my memory dredged that from or if it is correct. Please let me know the real answer.)

Doxygen -u Command Changes
----------------------------------------
I also ran the doxygen upgrade command on the Doxyfile.in, so that it reflects the latest valid doxygen settings. This made a lot of changes the comments in the Doxyfile.in, to bring them up to date with the latest doxygen usage. It added and removed some settings that you were not messing with anyway. So all of those changes are pretty insignificant.

The only changes I made was to include the build/examples/f03 directory, and to exclude directories below it which hold a bunch of cmake stuff we don't want in the documentation.

Code Changes
-------------------
There are almost no code changes in this pull request. In the C example code examplePio_c I added a check for the correct number of processors.

Old Subversion Keywords
--------------------------------
Many of the fortran files had old svn/cvs keywords (ex. $Version:$) in the comments at the top of the file. These have no meaning for git and were messing up the documentation, so I removed them.

Future Plans
----------------
I plan to continue to expand the example documentation as I gain understanding of the system. I wanted to get these changes merged, because many files were touched. Future development of the example documentation will hopefully involve only a few files.


CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 cmake_minimum_required (VERSION 2.8.12)
 project (PIO C Fortran)
 
+# The project version number.
+set(VERSION_MAJOR   2   CACHE STRING "Project major version number.")
+set(VERSION_MINOR   3   CACHE STRING "Project minor version number.")
+set(VERSION_PATCH   8   CACHE STRING "Project patch version number.")
+mark_as_advanced(VERSION_MAJOR VERSION_MINOR VERSION_PATCH)
+
```
jedwards4b: The current version (tag) in git is 2_0_26 (there is some legacy reason that i've long forgotten for using _ instead of . we can probably change that)  Is there a way to automatically have that git tag inserted here?

edhartnett: Strongly suggest you use . instead of _ in the version number. It will make
distributions much easier.

The best end-state is that the ParallelIO library is made available on
standard Linux distributions. This is already the case for netCDF-4 and
HDF5, compiled with mpicc and built with MPICH. So it's possible to set up
the full parallel development environment for netCDF-4 using only yum. Very
handy for users and developers! Yum-based installs prevent support emails
asking about the build (which *is* a tricky build, with lots of mistake
that can be made along the way.)

Of course, on HPC there will always be a lot of building from source, but
programmer's workstations are usually Linux.

The closer we adhere to the standard Linux distribution conventions, the
easier it is to get others to help distribute the library.

Thanks,
Ed

PR #39, opened 2015-12-10
precount outstanding requests so that request and status are sized pr…

jedwards4b:
…operly

### DIFF
--- a/src/clib/pio_darray.c
+++ b/src/clib/pio_darray.c
@@ -1343,6 +1343,8 @@ int flush_output_buffer(file_desc_t *file, bool force, PIO_Offset addsize)
 #ifdef TIMING
   GPTLstart("PIO:flush_output_buffer");
 #endif
+  pioassert(file!=NULL,"file pointer not defined",__FILE__,__LINE__);
+
 
   ierr = ncmpi_inq_buffer_usage(file->fh, &usage);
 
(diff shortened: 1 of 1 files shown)


src/clib/pio_darray.c
```diff
@@ -1357,32 +1359,37 @@ int flush_output_buffer(file_desc_t *file, bool force, PIO_Offset addsize)
...
+      reqcnt+=vdesc->nreqs;
+      if(vdesc->nreqs>0) maxreq = i;
+    }
+    int request[reqcnt];
+    int status[reqcnt];
 
-
+    for(int i=0; i<=maxreq; i++){
+      vdesc = file->varlist+i;
 #ifdef MPIO_ONESIDED
```
edhartnett: What is MPIO_ONESIDED?

src/clib/pio_darray.c
```diff
@@ -1357,32 +1359,37 @@ int flush_output_buffer(file_desc_t *file, bool force, PIO_Offset addsize)
...
+    int reqcnt;
+    maxreq = 0;
+    reqcnt=0;
+    rcnt=0;
     for(int i=0; i<PIO_MAX_VARS; i++){
       vdesc = file->varlist+i;
+      reqcnt+=vdesc->nreqs;
+      if(vdesc->nreqs>0) maxreq = i;
+    }
+    int request[reqcnt];
```
edhartnett: Wow I didn't even know you could do that in C. I would have thought that a malloc was required. Didn't know you could declare an array with a size you just computed!

src/clib/pio_darray.c
jedwards4b: Early this year I worked on IO performance with ROMIO (MPI-IO) and PnetCDF developers.
The MPIO_ONESIDED optimization went into the ROMIO library - that optimization should eventually make it into the mpich distribution but as far as I know that hasn't happened yet.  
It improves performance but introduces restrictions on what you can put into the write buffer - 
the buffer (aggregated across IO tasks) must represent a contiguous block on the file.  This ifdef is about trying to assure that that is the case.

src/clib/pio_darray.c
jedwards4b: It's part of the c99 standard.

src/clib/pio_darray.c
edhartnett: OK, so I should make this settable from cmake, as I did for PIO_USE_MALLOC?

I will take care of it in the next merge...

PR #54, opened 2016-02-08
Fix lengths of names of atts/vars/dims sent to tasks where rank != 0 for serial io types

edhartnett:
In current code, the null terminator is not broadcast to tasks where rank != 0 for serial formats on PIOc_inq_varname/dimname/attname.

### DIFF
--- a/src/clib/ncparser.pl
+++ b/src/clib/ncparser.pl
@@ -301,7 +301,7 @@ print F
 		  print F "    if(name != NULL){ \n";
 		  print F "      int slen;\n";
 		  print F "      if(ios->iomaster)\n";
-		  print F "        slen = (int) strlen(name);\n";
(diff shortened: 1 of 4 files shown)


jedwards4b:
pio_nc.c is generated from file pio_c_template.c using ncparser.pl.
The correct place to make this change is around line 262 of ncparser.pl 
The new pio_nc.c file is then generated by running
perl ncparser.pl $PNETCDF_INC_PATH $NETCDF_INC_PATH

make sure to use pnetcdf 1.6.1 or newer and netcdf 4.3.3.1 for the locations of the include directories.

edhartnett:
OK, thanks for pointing that out! 

I have changed the branch to reflect this. 

Thanks,
Ed

PR #56, opened 2016-02-09
Bug fixes, documentation improvement, build flags

edhartnett:
# NetCDF-4 Bug FIxes

Based on recent discussions with Jim I now have a much better understanding and I see that there is a bug in some of the nc4 functions. They do not broadcast their results to the other tasks, and they need to. This pull request includes the necessary fixes for this.

# Documentation

When I put the build instructions in the documentation, somehow I left out the netCDF fortran library. I have added build instructions for that and updated to the latest versions as well. 

# Build Flags

The c99 build flags are now added in the tests/unit directory. After I clear the current bug I will look into moving these flag definitions to a common place, since they seemingly will be needed in every directory with C code. But this pull request should get cdash passing again.

# Testing

This branch is building cleanly under Jenkins for me.


jedwards4b:
You have some lines reversed in pio_nc4.c :
errstr is intended to pass the error message to the check_netcdf handler and so should be allocated and written before the call.

Also I'm not sure I like the 
   return PIO_EIO 
lines since they introduce a memory leak because errstr is not deallocated - can these just set ierr and allow control to flow to the end of the subroutine?

edhartnett:
OK, good catches, thanks!

I have fixed and pushed changes to branch.

Thanks,
Ed

src/clib/pio_nc4.c
```diff
@@ -78,12 +86,20 @@ int PIOc_def_var_deflate(int ncid, int varid, int shuffle, int deflate,
...
+    {
 	errstr = (char *) malloc((strlen(__FILE__) + 20)* sizeof(char));
 	sprintf(errstr,"in file %s",__FILE__);
     }
-    if(errstr != NULL) free(errstr);
+
+    /* Check the netCDF return code, and broadcast it to all tasks. */
+    ierr = check_netcdf(file, ierr, errstr,__LINE__);
+
+    /* Free the error string if there is no error. */
```
jedwards4b: the comment should read free the errstr if it is allocated

edhartnett:
I have put the comment changes you suggest on to my next branch so they
will eventually make it to master. ;-)

Thanks,
Ed

PR #59, opened 2016-02-29
Fortran implementation and tests of netCDF-4 performance functions

edhartnett:


src/clib/pio_nc.c
```diff
@@ -3483,6 +3483,7 @@ int PIOc_inq_dimlen (int ncid, int dimid, PIO_Offset *lenp)
     sprintf(errstr,"in file %s",__FILE__);
   }
   ierr = check_netcdf(file, ierr, errstr,__LINE__);
+  printf("sizeof(MPI_OFFSET)=%d sizeof(PIO_Offset)=%d sizeof(*lenp)=%d \n", sizeof(MPI_OFFSET), sizeof(PIO_Offset), sizeof(*lenp));
```
jedwards4b: is this a debug statement you meant to remove?

src/clib/pio_nc4.c
```diff
@@ -1023,49 +844,42 @@ int PIOc_get_chunk_cache(int iotype, int io_rank, size_t *sizep,
...
-    /* if (sizep) */
-    /* 	if ((ierr = MPI_Bcast(sizep, 1, MPI_UNSIGNED_LONG, ios->ioroot, ios->my_comm))) */
-    /* 	    return PIO_EIO; */
-    /* if (nelemsp) */
-    /* 	if ((ierr = MPI_Bcast(nelemsp, 1, MPI_UNSIGNED_LONG, ios->ioroot, ios->my_comm))) */
-    /* 	    return PIO_EIO; */
-    /* if (preemptionp) */
-    /* 	if ((ierr = MPI_Bcast(preemptionp, 1, MPI_FLOAT, ios->ioroot, ios->my_comm))) */
-    /* 	    return PIO_EIO; */
+    if (ierr)
```
jedwards4b: I think that this is incorrect and may cause a hang on error If some tasks set ierr but others do not.  The check_netcdf function handles this correctly.

src/clib/pio_nc4.c
```diff
@@ -1112,6 +926,11 @@ int PIOc_set_var_chunk_cache(int ncid, int varid, size_t size, size_t nelems,
     ios = file->iosystem;
     msg = PIO_MSG_SET_VAR_CHUNK_CACHE;
 
+    int my_rank;
+    int ret;
+    if ((ret = MPI_Comm_rank(MPI_COMM_WORLD, &my_rank)))
```
jedwards4b: Again if MPI_Comm_rank fails on some but not all tasks - what will happen here?

src/clib/pio_nc4.c
```diff
@@ -1206,6 +1023,11 @@ int PIOc_get_var_chunk_cache(int ncid, int varid, size_t *sizep, size_t *nelemsp
 	return PIO_EBADID;
     ios = file->iosystem;
 
+    int my_rank;
+    int ret;
+    if ((ret = MPI_Comm_rank(MPI_COMM_WORLD, &my_rank)))
```
jedwards4b: same as above

tests/unit/ncdf_tests.F90
```diff
@@ -277,29 +277,96 @@ Subroutine test_nc4(test_id, err_msg)
...
+    real :: chunk_cache_preemption_in
+
     err_msg = "no_error"
 
     dims(1) = 2*ntasks
     compdof = 2*my_rank+(/1,2/)  ! Where in the global array each task writes
     data_to_write = 1+my_rank
 
-    print*, ''
+    print*, 'PIO_initdecom'
```
jedwards4b: spelling

jedwards4b:
review complete - a few minor issues

src/clib/pio_nc.c
edhartnett: Good catch. Fixed.

src/clib/pio_nc4.c
edhartnett: Those were for debugging purposes and should have been removed. They are removed now.

src/clib/pio_nc4.c
edhartnett: Removed.

tests/unit/ncdf_tests.F90
edhartnett: Fixed.

src/clib/pio_nc4.c
edhartnett: I can't find this code...

PIOC_get/set_chunk_cache cannot use check_netcdf, which depends on a file being specified. The chunk cache functions are called before a file is opened, to set the chunk cache for subsequently opened files. So in those functions, I use MPI_BCAST to send the error to all processes. This is tested in the test code, and all processes do receive the error code...

jedwards4b: Suppose you are using NETCDF4C and the call from io-task 0 returns an error, all other tasks have ierr=PIO_NOERR,
what happens here?

edhartnett: OK, good question!

I have changed code to handle this better, and also added a test which
passes in a deliberately bad value for preemption to ensure that all tasks
get the resulting error code.

Please take a look at the changes and let me know what you thiink.

Thanks,
Ed

edhartnett:
Thanks!

PR #61, opened 2016-03-18
now duplicate MPI communicators in init_intracomm

edhartnett:


edhartnett:
...I also added some comments to the function. Please take a look and let's ensure that I really understand what is going on. ;-)

PR #64, opened 2016-05-04
added documentation to iosystem_desc_t

edhartnett:


src/clib/pio.h
```diff
@@ -139,43 +139,89 @@ typedef struct io_desc_t
...
+  /** This is an MPI intra communicator that includes all the tasks
+   * involved in computation. */
   MPI_Comm comp_comm;
+
+  /** This is an MPI inter communicator between IO communicator and
+   * computation communicator. */
   MPI_Comm intercomm;
+
+  /** This is a copy (but not an MPI copy) of either the comp (for
+   * non-async) or the union (for async) communicator. */
```
jedwards4b: It should probably be an mpi copy, good point.

src/clib/pio.h
```diff
@@ -139,43 +139,89 @@ typedef struct io_desc_t
...
+
+  /** The rank of this process in the computation communicator, or -1
+   * if this process is not part of the computation communicator. */
   int comp_rank;
+
+  /** The rank of this process in the IO communicator, or -1 if this
+   * process is not part of the IO communicator. */
   int io_rank;
 
-  bool iomaster;
```
jedwards4b: So I guess I'm confused - these were bool but not anymore?

src/clib/pio.h
```diff
@@ -139,43 +139,89 @@ typedef struct io_desc_t
...
   int comproot;
+
+  /** An array of the ranks of all IO tasks within the union
+   * communicator. */
   int *ioranks;
 
+  /** Controls handling errors. */
   int error_handler;
+
+  /** ??? */
```
jedwards4b: Is this used anyplace - if not let's remove it.

src/clib/pio.h
jedwards4b: Nevermind - this should be a copy of the pointer not a distinct object.

PR #66, opened 2016-05-04
changing in pio_file.c to support async

edhartnett:
These changes do not affect non-async operation, but are necessary for async.

The changes are mostly sending the contents of the parameters to the io tasks, and sending back the resulting information, if any.

Some new MSG constants are added to support some missing functions in the original list.

### DIFF
--- a/src/clib/pio_file.c
+++ b/src/clib/pio_file.c
@@ -417,7 +417,7 @@ int PIOc_sync (int ncid)
   msg = PIO_MSG_SYNC;
 
   if(ios->async_interface && ! ios->ioproc){
-    if(!ios->comp_rank) 
+    if(ios->comp_rank == 0) 
       mpierr = MPI_Send(&msg, 1,MPI_INT, ios->ioroot, 1, ios->union_comm);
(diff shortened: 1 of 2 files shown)


src/clib/pio_file.c
```diff
@@ -405,9 +417,9 @@ int PIOc_sync (int ncid)
   msg = PIO_MSG_SYNC;
 
   if(ios->async_interface && ! ios->ioproc){
-    if(ios->compmaster) 
+    if(!ios->comp_rank)
```
jedwards4b: ! ios->comp_rank is confusing - isn't ios->compmaster the boolean for this?  Otherwise use ios->comp_rank == 0.

edhartnett: In the MPI C API, compmaster is not a boolean, it is an integer, set to either 0 or MPI_ROOT.

I have changed the if statement as you suggest.

jedwards4b: I don't see it here - did you push your change? (Push your change to the same branch to update the PR)

edhartnett: It's there now, take another look.

PR #67, opened 2016-05-05
changes to pioc.c, pio_lists.c to support async, also some temporary code

edhartnett:
In this pull request there are changed for pioc.c, mostly in the finalize function. The changes to compmaster and iomaster are in accordance with MPI docs and pass all tests. 

There is also a change to pio_lists.c, which used to change the ncids when run with async. That doesn't work with the async code I wrote, so I have taken it out and cleaned up the function, as well as addind documentation.

The pull request also includes copies of three code files, which will be used for async development. The original code files are generated by perl scripts. After async development is complete, I will modify the perl scripts to output the async-capable code, and remove these three files. For now I would like to merge them to master before I make async changes, so you can easily see the changes that I make.

### DIFF
--- a/src/clib/pioc.c
+++ b/src/clib/pioc.c
@@ -386,8 +386,8 @@ int PIOc_Init_Intracomm(const MPI_Comm comp_comm,
       iosys->intercomm = MPI_COMM_NULL;
       iosys->error_handler = PIO_INTERNAL_ERROR;
       iosys->async_interface= false;
-      iosys->compmaster = false;
-      iosys->iomaster = false;
+      iosys->compmaster = 0;
+      iosys->iomaster = 0;
       iosys->ioproc = false;
       iosys->default_rearranger = rearr;
       iosys->num_iotasks = num_iotasks;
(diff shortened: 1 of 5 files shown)


src/clib/pioc.c
```diff
@@ -500,27 +501,38 @@ int PIOc_set_hint(const int iosysid, char hint[], const char hintval[])
...
-   */ 
-  if(ios->intercomm == MPI_COMM_NULL){
-    if(ios->ioranks != NULL){
-      free(ios->ioranks);
-    }
+    return PIO_EBADID;
+  
+  /* If asynch IO is in use, send the PIO_MSG_EXIT message from the
+   * comp master to the IO processes. */
+  if (ios->async_interface && !ios->comp_rank)
```
jedwards4b: I find !ios->comp_rank a confusing way to point to comp rank 0.  I wonder if we should introduce a constant - say COMP_MASTER_RANK set to 0 so that we have ios->comp_rank == COMP_MASTER_RANK ?   Do the same for IO_MASTER_RANK?

edhartnett: I have changed it to ios->comp_rank == 0.

PR #69, opened 2016-05-26
Ejh async2

edhartnett:
Changes in support of async writes with darrays.

As discussed in meeting on 5/25/16 at Mesa Lab.


src/clib/pio_file.c
```diff
@@ -113,35 +147,45 @@ int PIOc_openfile(const int iosysid, int *ncidp, int *iotype,
...
-	// reset file markers for NETCDF on all tasks
-	file->iotype = PIO_IOTYPE_NETCDF;
-
-	// open netcdf file serially on main task
-        if(ios->io_rank==0){
-	  ierr = nc_open(filename, file->mode, &(file->fh)); }
-
-    }
-#endif
+/* #ifdef _NETCDF */
```
jedwards4b: Thought that we decided to leave this in place?

src/clib/pio_msg.c
```diff
@@ -0,0 +1,2179 @@
...
+	return PIO_EIO;
+    LOG((2, "inq_attid_handler got ncid = %d varid = %d attnum = %d id_present = %d",
+	 ncid, varid, attnum, id_present));
+
+    /* Match NULLs in collective function call. */
+    if (id_present)
+	idp = &id;
+
+    /* Call the function to learn about the attribute. */
+    if ((ret = PIOc_inq_attid(ncid, varid, name, idp)))
```
jedwards4b: Don't you need to broadcast the newly acquired idp back to the compute tasks?

src/clib/pio_msg.c
```diff
@@ -0,0 +1,2179 @@
...
+    LOG((1, "att_get_handler ncid = %d varid = %d namelen = %d name = %s iotype = %d"
+	 "atttype = %d attlen = %d typelen = %d",
+	 ncid, varid, namelen, name, iotype, atttype, attlen, typelen));    
+
+    /* Allocate space for the attribute data. */
+    if (!(ip = malloc(attlen * typelen)))
+	return PIO_ENOMEM;
+    
+    /* Call the function to read the attribute. */
+    if ((ierr = PIOc_get_att(ncid, varid, name, ip)))
```
jedwards4b: ip needs to be sent back to compute tasks

src/clib/pio_msg.c
```diff
@@ -0,0 +1,2179 @@
...
+	if (!(data = malloc(size_in_bytes)))
+	    return PIO_ENOMEM;
+	if ((mpierr = MPI_Bcast(data, size_in_bytes, MPI_INT, 0,
+				ios->intercomm)))
+	    return PIO_EIO;
+	LOG((2, " def_var_handler got parameters namelen = %d "
+	     "name = %s len = %d ncid = %d", namelen, name, len, ncid));
+
+	/* Call the create file function. */
+	if ((ret = PIOc_put_vara_int(ncid, varid, start, count, data)))
```
jedwards4b: Function name suggest read or write, but there is only write here.

src/clib/pio_file.c
edhartnett: I have uncommented these lines of code again, as you request.

My intention is NOT to do this with the new PIOc_open call. If this kind of format change is going to take place, it should be in the user's code, where they take full responsibility for the format they are using.

If we bury netcdf or data bugs under code like this, then they will never get fixed.

src/clib/pio_msg.c
edhartnett: The id is bcast at the end of PIOc_int_attid. None of the msg functions have to bcast their results, that is always done by the PIOc_ functions.

src/clib/pio_msg.c
edhartnett: Yep. ;-)

BTW all of these functions are at least minimally tested in test_intercomm.c. I'll add more extensive testing over the months to come.

src/clib/pio_msg.c
edhartnett: This function should have been (and now is) removed. It was an artifact of an earlier attempt in which I intended to use one handler for puts and gets. But the code between the two was mostly different so I use two separate functions now.

PR #96, opened 2016-08-16
wrappers for nc_open/nc_create

edhartnett:
Here are the wrappers for nc_open and nc_create. They use the same code behind the scenes as PIOc_openfile() and PIOc_createfile() but with a few twists.

An advantage to using these new wrapper functions is that it transparently enables access to the CDF5 format from Argonne parallel-netcdf, even for netCDF sequential.

### DIFF
--- a/src/clib/pio.h
+++ b/src/clib/pio.h
@@ -45,24 +45,24 @@
  */
 typedef struct var_desc_t
 {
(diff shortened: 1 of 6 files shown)


jedwards4b:
looks good, thanks

Issue #99, opened 2016-08-26
Fixing "duplicate" files in PIO global file list


jayeshkrishna:
PIO uses a global file list, pio_file_list (pio_lists.c), to keep track of 
files. This list is currently indexed using the file id (file handle) and 
the current code implicitly assumes that file handles for files are
unique globally (across all MPI processes).
However this assumption is not true when performing file operations
across multiple communicators (PIO iosystems). This issue results
in duplicate (same file id) file_desc_t entries in the global file list
in certain scenarios.

jayeshkrishna:
The jayeshkrishna/pio2/fix_dup_files_in_pio_file_list branch has two tests (pio_iosystem_test[2]) that show the issues resulting from the assumption above.

jayeshkrishna:
One of the issues is already fixed in the branch ( jayeshkrishna/pio2/fix_dup_files_in_pio_file_list ).
The overall solution that I propose (to fix all issues) is to index the global PIO file list using both file id and the communicator. This involves changing the C interface of most of the PIO functions (minor change but touches a lot of code).
@jedwards4b / @edhartnett / @Katetc : Please let me know your thoughts.

jayeshkrishna:
On the branch ( jayeshkrishna/pio2/fix_dup_files_in_pio_file_list ), pio_iosystem_test.F90 passes (the branch has a fix for the issue identified by this test) and pio_iosystem_test2.F90 fails/hangs due to the issue that I have mentioned above.

edhartnett:
I think the answer is to retain the ncids provided by the underlying libraries, but to add an internal ncid that is assigned to each ncid/communicator combination. This achieves the indexing you desire.

Then there need be no change of the C API. The user will still keep track of things with an ncid.

jedwards4b:
I'm not sure I understand why this would change the c interface - wouldn't it just change the calculation of the fh value based on communicator which is already available inside the routines?

jayeshkrishna:
IMO, changing the C interface to allow the fortran interface to pass the iosysid should allow us to search to search the list with ncid+iosysid. Adding another internal ncid might be complicating the design.

edhartnett:
Adding another internal ncid is necessary for other reasons. The parallel-netcdf library and the netCDF classic library will start overlapping ncids after the first 3 parallel-netcdf files are open. (parallel-netcdf uses 0 for it's first ncid, then increments. NetCDF classic starts with 3.)

[...]

PR #100, opened 2016-09-02
Fix dup files in pio file list - part 1

jayeshkrishna:
PIO uses a global file list, pio_file_list (pio_lists.c), to keep track of
files. This list is currently indexed using the file id (file handle) and
the current code implicitly assumes that file handles for files are
unique globally (across all MPI processes).

This commit fixes one part of the problem mentioned in Issue #99 .

This commit also brings in two tests, pio_iosystem_tests.F90 and
pio_iosystem_tests2.F90 . The first test passes now but the
second one fails/hangs.

### DIFF
--- a/src/clib/pio_file.c
+++ b/src/clib/pio_file.c
@@ -29,10 +29,11 @@
 int PIOc_openfile_retry(const int iosysid, int *ncidp, int *iotype,
 			const char *filename, const int mode, int retry)
 {
-    iosystem_desc_t *ios;  /* Pointer to io system information. */
-    file_desc_t *file;     /* Pointer to file information. */
-    int ierr = PIO_NOERR;  /* Return code from function calls. */
-    int mpierr = MPI_SUCCESS, mpierr2;  /* Return code from MPI function codes. */
+    iosystem_desc_t *ios;  /** Pointer to io system information. */
+    file_desc_t *file;     /** Pointer to file information. */
+    int tmp_fh = -1;
+    int ierr = PIO_NOERR;  /** Return code from function calls. */
+    int mpierr = MPI_SUCCESS, mpierr2;  /** Return code from MPI function codes. */
 
     LOG((1, "PIOc_openfile iosysid = %d", iosysid));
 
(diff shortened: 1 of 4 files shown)


jayeshkrishna:
The branch in this PR - jayeshkrishna/pio2/fix_dup_files_in_pio_file_list_merge_master - is a merge of jayeshkrishna/pio2/fix_dup_files_in_pio_file_list with a latest branch from master. This new branch *_merge_master was created to ease (automatically merge) the merging (the conflict was minor).

src/clib/pio_file.c
```diff
@@ -200,8 +203,14 @@ int PIOc_openfile_retry(const int iosysid, int *ncidp, int *iotype,
 	if ((mpierr = MPI_Bcast(&file->mode, 1, MPI_INT, ios->ioroot, ios->union_comm)))
 	    return check_mpi(file, mpierr, __FILE__, __LINE__);
       
-	if ((mpierr = MPI_Bcast(&file->fh, 1, MPI_INT, ios->ioroot, ios->union_comm)))
+      tmp_fh = file->fh;
+	if ((mpierr = MPI_Bcast(&tmp_fh, 1, MPI_INT, ios->ioroot, ios->union_comm)))
```
edhartnett: Seems like tmp_fh is not needed. Why not just Bcast &file->fh?

jayeshkrishna:
The file->fh is set by the I/O library (pnetcdf/netcdf) and is set by PIO only if needed (if not set by the underlying I/O library).

edhartnett:
Yes I understand that file->fh needs to be set on all processes because the pnetcdf is not doing it. (I believe netCDF will correctly return ncid on all processes - please tell me if I am wrong and we can get that fixed).

But the same could be accomplished without introducing the new variable tmp_fh. Instead, just use file->fh directly in the Bcast, and remove tmp_fh.

jayeshkrishna:
Also note that file->fh does not have to be the same across all processes (active I/O processes).

edhartnett:
I'm sorry, but I don't understand if you are responding to my comment.

My point is that the tmp_fh variable can be removed. Unneeded variables only clutter up the code. Can you change your pull request so that you set file->fh directly, and don't use tmp_fh at all?

Thanks!

src/clib/pio_file.c
jedwards4b: tmp_fh is used just below at line 212.  In pnetcdf or netcdf4p the filehandle is set on all iotasks and may differ from one task to another - this mechanism prevents it from being overwritten.

PR #101, opened 2016-09-07
a rewrite of the async code which initializes the communicators...

edhartnett:
This pull request contains a complete rewrite of the way that the async code creates communicators.

In the previous implementation (patterened after PIO1), the intercomm was created using the world as the communicator in which all processes can be found. This did not work in the case of multiple computational components, because it required that all processes participate in the construction of the intercomm.

In the new implementation, the union comm is used as the communicator in which all processes can be found for the intercomm. Since the union comm contains only the IO procs and the procs involved in that computation component, the intercomm can be created without any other processes being involved.

I have had to change the initialization of the system to this:
    int PIOc_Init_Async(MPI_Comm world, int num_io_procs, int *io_proc_list, int component_count,
			int *num_procs_per_comp, int **proc_list, int *iosysidp);

The existing Init_Intercomm will be removed.

I have added a new test, test_intercomm3, which at the moment is much like test_intercomm2, but which will become more complex and different as I add more testing for multiple computation components.


src/clib/pio_msg.c
```diff
@@ -1616,61 +1645,602 @@ int pio_msg_handler(int io_rank, int component_count, iosystem_desc_t *iosys)
...
+	if (open_components == 0)
+	    break;
+    }
+
+    LOG((3, "returning from pio_msg_handler"));
+    return PIO_NOERR;
+}
+
+/* int */
+/* pio_iosys_print(int my_rank, iosystem_desc_t *iosys) */
```
jedwards4b: Should this be removed?

src/clib/pio_msg.c
```diff
@@ -1616,61 +1645,602 @@ int pio_msg_handler(int io_rank, int component_count, iosystem_desc_t *iosys)
...
+	 comp_comm, peer_comm, cmp));
+
+    /* Set these values for all comp comms. */
+    my_iosys->io_comm = MPI_COMM_NULL;
+    my_iosys->iogroup = MPI_GROUP_NULL;
+    my_iosys->io_rank = -1;
+    my_iosys->ioproc = false;
+    my_iosys->iomaster = MPI_PROC_NULL;
+
+    /* If multiple components are in use, this function will be called
```
jedwards4b: I'm not sure about this comment - sometimes components overlap tasks and so comp_comm can have values on more than one pass.

src/clib/pio_msg.c
```diff
@@ -1802,295 +2372,433 @@ int PIOc_Init_Intercomm(int component_count, MPI_Comm peer_comm,
...
+    /* 	    ierr = check_mpi(NULL, mpierr, __FILE__, __LINE__); */
+
+    return ierr;
+}
+
+/** Set up the communicators for parallel I/O.
+ *
+ * @param world the communicator containing all the available tasks.
+ * @param component_count number of computational components
+ * @param color for this task. Use 0 for an I/O task, 1 for the first
```
jedwards4b: I think that you are assuming each task is in only one computational group.  This is not correct.

jedwards4b:
It seems like the comments imply that there is only zero or one comp comm per task - but I didn't see anything in the code which made that assumption.    The comments for PIOc_Init_async are completely wrong and there is a lot of commented out code, can you make a clean up pass before we merge?

src/clib/pio_msg.c
edhartnett: Umm, the way I have written this code, components can not overlap. All components (including the IO component) must be disjoint.

jedwards4b: In that case I cannot accept this PR - compute components should be allowed to overlap.  But I'm not sure I can find a spot in the code which relies on this assumption.

edhartnett: Stand by while I work on this a bit.

Can you explain the use case for having a processor involved in more than one component? Why would you want to do that?

Thanks,
Ed

jedwards4b: Yes, it's actually pretty common - In cesm atm needs boundary conditions from cice and lnd cice and lnd can run concurrently but only after atm has completed a step, atm cannot compute the next step until it gets results from cice and lnd so we generally will have cice and lnd use a subset of atm tasks - so for example atm uses tasks 1-640, lnd uses tasks 1-320 and cice uses 321-640

edhartnett: OK, I think I can accommodate that. Let me add a test for it...

jayeshkrishna:
@edhartnett / @jedwards4b : When (estimate) will these changes be merged to master? Also, will the non async versions (pio_darray.c vs pio_darray_async.c) be deleted once this merge is complete?

edhartnett:
I would like to merge these changes now, and continue working on a new branch.

For the moment async initializations with more than one computational component are disallowed in the code, while I work on a bug finalizing when more than one computational component is in use.

The issue of overlapping computational components is not yet handled but will be in my next branch.

These code changes do not affect the non-async operation of the PIO library.

tests/general/CMakeLists.txt
```diff
@@ -348,20 +348,20 @@ else ()
...
-#         NUMPROCS 4
-#         TIMEOUT ${DEFAULT_TEST_TIMEOUT})
-# endif ()
+if (PIO_USE_MPISERIAL)
+    add_test(NAME pio_iosystem_tests2
+        COMMAND pio_iosystem_tests2)
+    set_tests_properties(pio_iosystem_tests2
+        PROPERTIES TIMEOUT ${DEFAULT_TEST_TIMEOUT})
+else ()
+    # add_mpi_test(pio_iosystem_tests2
```
jayeshkrishna: Don't turn off this test (and merge to master). It is known to hang/fail and will timeout. I am going to fix this once you finish merging this branch (so that my changes are easier to merge).

edhartnett: OK I will turn it back on...

jayeshkrishna: Thanks

edhartnett: When you check in tests that fail, it breaks my continuous integration testing.

(That is, I have a jenkins task that is building the code every time it changes, so that I can see if I break something. If you check in a test that fails, then it shows as broken every time, defeating the purpose of continuous integration.)

At the moment let's just leave this as it is until I get my current PR merged. Then I would like to turn off the test on master again, so that I can make the changes that we discussed WRT fully merging in the async code so that there are no longer two different versions of some PIO functions.

But the test will still be there, and as soon as I am done, you can take a branch, turn the test back on in the branch, and proceed with your fix.

jayeshkrishna: I don't think it breaks the idea of CI, the code has an existing bug that is exposed by this test. Doesn't the output of your tests show which test failed?

jayeshkrishna: IMO, there is no advantage of turning off a failing test

src/clib/pio_msg.c
```diff
@@ -1600,80 +1680,88 @@ int pio_msg_handler(int io_rank, int component_count, iosystem_desc_t *iosys)
...
-	   my_rank, iosys->ioroot, iosys->comproot, iosys->iomaster,
-	   iosys->compmaster);
-
-    printf("%d iotasks:", my_rank);
-    for (int i = 0; i < iosys->num_iotasks; i++)
-	printf("%d ", iosys->ioranks[i]);
-    printf("\n");
-    return PIO_NOERR;
-}
+/* int */
```
jedwards4b: Should we remove this instead of leaving it in comments?

tests/unit/test_async_2comp.c
```diff
@@ -0,0 +1,127 @@
...
+		ERR(ret);
+
+    	    /* Check the file for correctness. */
+    	    if ((ret = check_nc_sample_1(iosysid[my_comp_idx], flavor[flv], filename, my_rank)))
+    	    	ERR(ret);
+
+    	} /* next netcdf format flavor */
+
+	/* If I don't sleep here for a second, there are problems. */
+	sleep(2);
```
jedwards4b: Why is that - maybe a MPI_barrier instead of a sleep?

edhartnett: This is a test that I am still working on. The sleep will certainly not remain in the code.

Right now this test is turned off. Once I get it working reliably I will turn it on again.

src/clib/pio_msg.c
edhartnett: Yes, I will remove this commented-out function. I wrote it before I put LOG into the library. Now that we have LOGs I don't use this function anymore anyway.

tests/general/CMakeLists.txt
edhartnett: Howdy @jayeshkrishna!

CI means that you can't leave broken tests turned on in the build, or else all CI builds show as broken.

The point of CI is that it is automated, not that I go look at what failed every time and say "that's OK." The idea is that I can tell at a glance if code I've just checked in caused any tests to fail.

My CI page builds the library in various configurations (ex with and without async, with and without logging, etc.), and so there are a number of builds that show up as broken when there is a failing test.

The way to track known bugs is in the github issues list, not by leaving in failing tests. 

There is a good description of the git workflow here: https://www.atlassian.com/git/tutorials/comparing-workflows/centralized-workflow (we are using what is called "Feature Branch Workflow")

From this document: 

> The core idea behind the Feature Branch Workflow is that all feature development should take place in a dedicated branch instead of the master branch. This encapsulation makes it easy for multiple developers to work on a particular feature without disturbing the main codebase. It also means the master branch will never contain broken code, which is a huge advantage for continuous integration environments.

We have to share the master branch, so we need to leave it in a working state at all times.

I have commented out this test on master. When you take your next branch, simply uncomment the test on your branch, and leave it commented out on master, so we can continue to use CI. When your merge your branch with the fix, you will also merge the uncommenting of the test, and so all will be right with the world.

Your fixes, and your working test, will be merged into master at the same time, so that master will always yield a stable build that passes all tests. 

Thanks!

rljacob: CIT is supposed to be a suite of tests and the whole thing shouldn't fail because one test failed.  You just have one test in the red column on Cdash and its stays red until you fix it.  It should also distinguish between a build fail and a runtime fail.   Its a good reminder.   Test Driven Design also encourages and allows failing tests.

jayeshkrishna:
@edhartnett / @jedwards4b : Thanks for merging this code into master, this should help me proceed with my work (Issue #99). I will re-enable the pio_iosystems_test2 test that @edhartnett disabled in my branch (I would still recommend that we no longer disable tests to get a 100% pass and instead leave the failing test as it is until we fix the bug).

edhartnett:
@jayeshkrishna there is still more code to merge. There is a current pull request up that @jedwards4b is reviewing. You probably want to wait until that is merged before taking your branch.

PR #103, opened 2016-09-17
complete merge of async code

edhartnett:
Removed the non-async version of the code from the build system and the code base.

This will make it easier for contributors to submit modifications.


CMakeLists.txt
```diff
@@ -14,7 +14,6 @@ mark_as_advanced(VERSION_MAJOR VERSION_MINOR VERSION_PATCH)
 #===== Library Options =====
 option (PIO_ENABLE_FORTRAN   "Enable the Fortran library builds"            ON)
 option (PIO_ENABLE_TIMING    "Enable the use of the GPTL timing library"    ON)
-option (PIO_ENABLE_ASYNC     "Enable the use of asychronus IO operations"   OFF)
```
jedwards4b: async code is now always enabled?

src/clib/pio_darray.c
```diff
@@ -226,8 +249,13 @@ int pio_write_darray_nc(file_desc_t *file, io_desc_t *iodesc, const int vid,
 #endif /* _NETCDF4 */
 	    case PIO_IOTYPE_NETCDF:
 	    {
+		/* Find the type size (again?) */
 		mpierr = MPI_Type_size(iodesc->basetype, &dsize);
+		
 		size_t tstart[ndims], tcount[ndims];
+
+		/* The IO master task does all the data writes, but
```
jedwards4b: When using the serial netcdf PIO acts like a funnel.   Data is moved from compute tasks to IO tasks then from IO tasks to IO task 0 for write.  For read data the opposite happens, data is read on IO task 0 distributed to all io tasks and then to compute tasks.   We could optimize these data paths but serial netcdf is not the primary mode for PIO.   Can you change this comment?

src/clib/pio_darray.c
```diff
@@ -372,50 +397,65 @@ int pio_write_darray_nc(file_desc_t *file, io_desc_t *iodesc, const int vid,
...
+ * @return 0 for success, error code otherwise.
+ * @ingroup PIO_write_darray
  */
-int pio_write_darray_multi_nc(file_desc_t *file, const int nvars, const int vid[],
-			      const int iodesc_ndims, MPI_Datatype basetype, const PIO_Offset gsize[],
+int pio_write_darray_multi_nc(file_desc_t *file, const int nvars, const int *vid,
+			      const int iodesc_ndims, MPI_Datatype basetype, const PIO_Offset *gsize,
 			      const int maxregions, io_region *firstregion, const PIO_Offset llen,
 			      const int maxiobuflen, const int num_aiotasks,
-			      void *IOBUF, const int frame[])
```
jedwards4b: Is there a difference between const int frame[] and const int *frame?  Personal preference?

CMakeLists.txt
edhartnett: Correct.

src/clib/pio_darray.c
edhartnett: Strong C convention.

src/clib/pio_darray.c
edhartnett: Done, thanks. Any updates to the comments are always welcome.

jedwards4b:
Merging this PR with the understanding that pio_iosystem_tests2 will be addressed next.

PR #112, opened 2016-10-13
More argument and return val checks

jayeshkrishna:
This commit adds more argument and return value checks in the
code. Also adding more description for some failure scenarios.

* Stricter argument value (null, invalid args) checks
* Checks for memory allocation errors (malloc/calloc/bget)

### DIFF
--- a/src/clib/pio_darray.c
+++ b/src/clib/pio_darray.c
@@ -40,8 +40,12 @@ PIO_Offset PIOc_set_buffer_size_limit(const PIO_Offset limit)
 {
     PIO_Offset oldsize;
     oldsize = PIO_BUFFER_SIZE_LIMIT;
-    if (limit > 0)
-	PIO_BUFFER_SIZE_LIMIT = limit;
+    if (limit > 0){
+      PIO_BUFFER_SIZE_LIMIT = limit;
+    }
+    else{
+      fprintf(stderr, "%s:%d ERROR: Invalid buffer size (%d) specified, unable to set buffer size limit\n", __FILE__, __LINE__, limit);
+    }
     return oldsize;
 }
 
@@ -117,6 +121,14 @@ int pio_write_darray_nc(file_desc_t *file, io_desc_t *iodesc, const int vid,
     GPTLstart("PIO:write_darray_nc");
 #endif
 
+    if((file == NULL) || (iodesc == NULL))
+      return PIO_EINVAL;
+
+    if(IOBUF == NULL){
+      fprintf("%s:%d ERROR: Trying to write NULL IOBUF\n", __FILE__, __LINE__);
+      return PIO_EINVAL;
+    }
+
     /* Get the IO system info. */
     if (!(ios = file->iosystem))
 	return PIO_EBADID;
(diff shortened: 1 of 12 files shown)


jayeshkrishna:
This PR should not change the behaviour of the code in any way. It adds more strict checks on argument values and return values from functions.

jayeshkrishna:
@edhartnett : Just an FYI, this PR touches a lot of code (although the changes are trivial args/retval checks)

jayeshkrishna:
I typically use assert() in internal functions (internal utility functions with less exposure to the user, only of interest to developer) where the conditions are not expected to fail, but if things fail we are done. Following the existing code style I used pioassert (assert+die in debug mode) for these cases and checks for others.

edhartnett:
I applaud the idea of additional arg checks and asserts. But I would like this PR to not be merged. Here's why:
* Should not be any print statements. Only return codes should be used to indicate error.
* Should use existing standard C assert which can be turned off with DEBUG flag on bcc and other compilers.
* Intersects with lots of code I am about to merge.
* Stylistic mismatch with existing code.

How about I take a stab at this after me next merge.

jayeshkrishna:
This PR would be the first step and we can work on other issues (error messages, C assert vs pioassert etc) in future PRs/Issues.
I prefer fixing things incrementally so that one can follow the changes (e.g. if we are replacing pioasserts with asserts it has to happen to all the code etc)

edhartnett:
OK this code was not up for review for a very long time. I believe we had agreed that in usual cases PRs would be up for a day before merging so that everyone can comment.

Merge of this PR should be undone for the following reasons:

* We have already decided that there should be no printed error statements in the library. Most of this PR is the addition of printf statements which will have to be removed. In some cases it is putting back in printfs where I had already removed them (after consulting with Jim and Kate).

* In many cases this code seems very cluttered, for example, when things can be checked within the list lookup function, they should be. There is no reason to put code in every function which checks that checks the file->ios pointer. If you want to check that, do it once in the function that finds the file info.

* There is no demonstrated need to introduce the pioassert function. The C library already includes an assert that is well understood and will do everything we need. Let us focus our code on what we want to achieve with parallel IO, and not attempt to re-write parts of the existing C library.

* By touching all these files you are making my merge of the ncid changes very difficult. There does not seem to be a pressing need to make my job harder here. Seems like putting in a bunch of error checking and asserts could have waited.

rljacob:
PIO should not print any error messages?

edhartnett:
PIO should not print any error messages.

That is up to the calling program. In the usual way of Unix libraries, PIO should return error codes and the calling program can handle it. We have also introduced a function which prints the error message that goes with the code.

jedwards4b:
Not printing error messages is more consistent with the behavior of netcdf/pnetcdf.

jayeshkrishna:
@edhartnett : You can work on this (please create a separate issue) if you want.

rljacob:
Checking return codes is always a good thing and PIO needs more of it.  PIO has not yet achieved the "never print error messages" standard so this code is in line with existing practice in the code.  I think this should be merged (again).

PR #113, opened 2016-10-13
Removing auto-generated test

jayeshkrishna:
A generated test was checked in, instead of the test template,
in 17634e98556cdf8ff2b7538cd6626d7a3ef0a3d0

This commit removes this generated test, pio_decomp_fillval2.F90,
and re-enables the original test, pio_decomp_fillval (that is
based on a test template)


jayeshkrishna:
The original test was removed in the commit mentioned above and was hiding a bug in the code. After re-enabling, pio_decomp_fillval, i see that it fails on master

PR #117, opened 2016-10-14
Fix fillvalue handling for netcdf

### LINKED ISSUE #114: pio_decomp_fillval test fails on master
The pio_decomp_fillval (pio_decomp_fillval.F90.in) fails on master (70c910bfa307108ac20d594a30698c089bb9dffb).

jayeshkrishna:
The code to handle fillvalues for NetCDF cases was commented out
in 93d430583fefbe2ef3c5f6407a7a4aea74e37301 . Uncommenting the
code so that fillvalues are handled correctly for NETCDF, NETCDF4P
and NETCDF4C .

Without this fix pio_decomp_fillval fails for NETCDF cases (the
test however passed for PNETCDF)

Fixes #114

### DIFF
--- a/src/clib/pio_darray.c
+++ b/src/clib/pio_darray.c
@@ -1135,29 +1135,27 @@ int PIOc_write_darray_multi(const int ncid, const int *vid, const int ioid,
 	switch(file->iotype)
 	{
 	case PIO_IOTYPE_PNETCDF:
+	case PIO_IOTYPE_NETCDF4P:
 	    ierr = pio_write_darray_multi_nc(file, nvars, vid,
 					     iodesc->ndims, iodesc->basetype, iodesc->gsize,
 					     iodesc->maxfillregions, iodesc->fillregion, iodesc->holegridsize,
 					     iodesc->holegridsize, iodesc->num_aiotasks,
 					     vdesc0->fillbuf, frame);
 	    break;
-	case PIO_IOTYPE_NETCDF4P:
 	case PIO_IOTYPE_NETCDF4C:
 	case PIO_IOTYPE_NETCDF:
-	    /*       ierr = pio_write_darray_multi_nc_serial(file, nvars, vid,
+	    ierr = pio_write_darray_multi_nc_serial(file, nvars, vid,
 		     iodesc->ndims, iodesc->basetype, iodesc->gsize,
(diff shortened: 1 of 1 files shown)


jayeshkrishna:
@jedwards4b : Do you recall why the code to handle fillvalues for NETCDF was commented out in 93d430583fefbe2ef3c5f6407a7a4aea74e37301 ?

pio_decomp_fillval test now passes (was FAILing for NETCDF cases) on this branch

src/clib/pio_darray.c
```diff
@@ -1135,29 +1135,27 @@ int PIOc_write_darray_multi(const int ncid, const int *vid, const int ioid,
...
-		     brel(vdesc0->fillbuf);
-		     vdesc0->fillbuf = NULL;
-		     }
-	    */
 	    break;
 	}
+    if (vdesc0->fillbuf != NULL){
+         /* printf("%s %d %x\n",__FILE__,__LINE__,vdesc0->fillbuf); */
+         brel(vdesc0->fillbuf);
+         vdesc0->fillbuf = NULL;
```
edhartnett: Some comments:
* Please remove the commented printf statement.
* Can you add a comment to the code explaining what is going on?
* Suggest you use strong C convention of if (vdesc0->fillbuff) instead of if (vdesc0->fillbuff != NULL).
* Also please conform to standard K&R style with respect to indentation and braces. Code that is consistent is easier to debug and maintain.

jayeshkrishna: Maybe we should do stylistic changes in a separate issue. This commit uncomments and moves around existing code (I want it to be clear what was done in this commit).

jedwards4b:
No I'm afraid I don't recall why this is commented.

src/clib/pio_darray.c
jedwards4b: I agree, I'm willing to agree on a certain coding style but we haven't had that discussion yet.

edhartnett: Even when I apply these changes, the test pio_decomp_fillval still fails for me on master.

jayeshkrishna: Works for me (on the branch and after merging to master, pio2_0_15-1068-g19df8fb). @jedwards4b : Are you able to get the pio_decomp_fillval test to pass after merging this branch to master?

jedwards4b:
This code was included in PR#122

PR #123, opened 2016-10-19
added code document

edhartnett:
As discussed I have added a short document describing code style for contributors to the C library. (I have left the Fortran library for another programmer who spends more time in that language.)

Best way to read this is to do a make doc and look at the HTML output.

The point of this document is not to teach about good C programming, but just to provide a very basic style guide so that the code looks consistent everywhere. A short (but complete) function from the code base is provided as an example.

This PR is documentation only, with no code changes.

### DIFF
--- a/doc/source/code_style.txt
+++ b/doc/source/code_style.txt
@@ -199,5 +199,11 @@ int PIOc_inq_attname(int ncid, int varid, int attnum, char *name)
 }
 </pre>
 
+## Further Information ##
+
+<p>For style issues not already covered in this document, see this <a
+href="https://www.kernel.org/doc/Documentation/CodingStyle">style
+guide</a>.
+
 _Last updated: 05-16-2016_
 */
--- a/doc/source/code_style.txt
+++ b/doc/source/code_style.txt
@@ -79,6 +79,24 @@ correctly. M-x indent-region will indent a selected region of code.
 
 <p>M-x untabify will convert all the tabs in a file to spaces.
 
+## Code Review ##
+
+<ul>
+
+<li>Pull requests will be as small as possible and focused on
+one(-ish) issue.
+
+<li>Pull requests will be left up on github for at least a day to
+allow all team members time to look at them.
+
+<li>Jim will identify key stakeholders in chaged code and ensure they
+accept code changes.
+
+<li>Reviewers are open-minded and ready to accept improvements to the
+library.
+
+</ul>
+
 ## Example ##
 
 <pre>
--- a/doc/source/code_style.txt
+++ b/doc/source/code_style.txt
@@ -39,12 +39,15 @@ add to it, carefully.
 <li>Assign and check return values in the same if statement.
 </ul>
 
-## Indentation ##
+## Indentation and Spacing##
 
 <ul>
 <li>Use spaces not tabs.
 <li>4 spaces is the unit of intendation.
 <li>Indentation as defined by the "linux" style in emacs (see below).
+<li>Use spaces around most operators (=+-*/) not pointer or prefix/postfile (*++--)
+<li>Spaces after most keywords (if, for, while, etc.)
+<li>No spaces after function name.
 </ul>
 
 ## Braces ##
--- a/doc/source/code_style.txt
+++ b/doc/source/code_style.txt
@@ -1,0 +1,182 @@
+/******************************************************************************
+ *
+ * 
+ *
+ * Copyright (C) 2016
+ *
+ * Permission to use, copy, modify, and distribute this software and its
+ * documentation under the terms of the GNU General Public License is hereby 
+ * granted. No representations are made about the suitability of this software 
+ * for any purpose. It is provided "as is" without express or implied warranty.
+ * See the GNU General Public License for more details.
+ *
+ * Documents produced by Doxygen are derivative works derived from the
+ * input used in their production; they are not affected by this license.
+ *
+ */ /*! \page code_style Code Style for Contributors
+
+# Introduction #
+
+This document describes the code style to be used when contributing to
+the PIO library.
+
+Consistency of style allows bugs to be spotted more quickly and
+contributes to code quality.
+
+# C Code #
+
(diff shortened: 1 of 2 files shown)


jayeshkrishna:
Why not use a widely accepted coding style/standard and customize it? For example (I am fine with any widely accepted style), start with https://www.kernel.org/doc/Documentation/CodingStyle and customize it, if needed, in a doc.

edhartnett:
That is what I did, this is the result.

jayeshkrishna:
I don't see a reference to any coding style (a document widely accepted and used by developers) in the document. We don't want to end up with a document that covers only some part of the coding style/guideline.

rljacob:
I was thinking of using the wiki here to write a PIO programmers guide and for style we can just point people to something like the Linux Kernel guide.  We don't have to actually write our own.

edhartnett:
I have updated the style guide as suggested.

PIO Programmers Guide is a different beast and I encourage you to write it. What we have here is a contributors style guide for C code, which many projects have.

Unless anyone has any further specific suggestions or objections I suggest we move on.

If there is desire for a similar style guide for Fortran, please feel free to add that.

doc/source/code_style.txt
```diff
@@ -0,0 +1,209 @@
...
+process.
+<li>Document in complete sentences.
+<li>Use C (not C++) comment delimiters.
+</ul>
+
+## Emacs ##
+
+<p>Put this in your .emacs file:
+
+<pre>
```
jayeshkrishna: Does this emacs setting change tabs to spaces?

doc/source/code_style.txt
```diff
@@ -0,0 +1,209 @@
...
+<p>The tab key (used anywhere on the line) will indent a line
+correctly. M-x indent-region will indent a selected region of code.
+
+<p>M-x untabify will convert all the tabs in a file to spaces.
+
+## Code Review ##
+
+<ul>
+
+<li>Pull requests will be as small as possible and focused on
```
jayeshkrishna: This information should be in a programmmer's guide (not style guide)

doc/source/code_style.txt
```diff
@@ -0,0 +1,209 @@
...
+            if ((mpierr = MPI_Bcast((void *)name, namelen + 1, MPI_CHAR, ios->ioroot,
+                                    ios->my_comm)))
+                check_mpi(file, mpierr, __FILE__, __LINE__);
+        }
+
+    return ierr;
+}
+</pre>
+
+## Further Information ##
```
jayeshkrishna: May be move this info to the top since this document is listing exceptions to this default style

doc/source/code_style.txt
```diff
@@ -0,0 +1,209 @@
...
+## Brevity ##
+
+<p>Consider: C is a terse language.
+
+<ul>
+<li>Use the fact that NULL == 0, and 0 == false, to shorten code.
+<li>Assign and check return values in the same if statement.
+</ul>
+
+## Indentation and Spacing##
```
jayeshkrishna: Is all the info below exceptions to the linux kernel C style ?

doc/source/code_style.txt
```diff
@@ -0,0 +1,209 @@
...
+<p>We cannot remove or change a function in the public C API. We can
+add to it, carefully.
+
+## Brevity ##
+
+<p>Consider: C is a terse language.
+
+<ul>
+<li>Use the fact that NULL == 0, and 0 == false, to shorten code.
+<li>Assign and check return values in the same if statement.
```
jayeshkrishna: I would prefer us checking the return values from a function in a new statement (IMO, makes the code easy to read)

doc/source/code_style.txt
edhartnett: No you must run untabify.

doc/source/code_style.txt
```diff
@@ -0,0 +1,209 @@
...
+<li>4 spaces is the unit of intendation.
+<li>Indentation as defined by the "linux" style in emacs (see below).
+<li>Use spaces around most operators (=+-*/) not pointer or prefix/postfile (*++--)
+<li>Spaces after most keywords (if, for, while, etc.)
+<li>No spaces after function name.
+</ul>
+
+## Braces ##
+
+<p>Put braces on their own line, avoiding their use if possible.
```
edhartnett: Yes, it's a deal-breaker. ;-)

Need them on separate lines to allow easy commenting out of if statement without having to hunt down other brace.

doc/source/code_style.txt
edhartnett: Mostly just a summary. Some exceptions.

doc/source/code_style.txt
edhartnett: Strong C convention indicates otherwise. No need to add those extra lines of code.

The idea of returning non-zero for errors was specifically built in to support this:

if (some_func())
    return ERROR;

Fortran I will leave to others. In C you check it in the if statement.

doc/source/code_style.txt
jayeshkrishna: ok

doc/source/code_style.txt
jayeshkrishna: It would be better to always convert tabs to spaces, IMO

doc/source/code_style.txt
rljacob: Lets say "we use linux kernel C style (link) with the exceptions below".  Saves us having to restate what we have in common.

doc/source/code_style.txt
edhartnett: What I have used on other projects, and would like to implement here, is a (voluntary) filter that is applied when code is checked in and applies formatting as automatically as possible.

In some environments (like eclipse) this is trivially easy to set up, but in emacs it will require a little messing about. But the idea is that everything that can be done automatically is done whenever a file is committed, and so the code in the repo is always meeting the standard.

So if you want to come up with an emacs function that runs untabify, and indents-region, that would be cool and we can add it to the document.

PR #126, opened 2016-10-19
check params - part 1

edhartnett:
In the recently rolled-back PR, the following code was being added to check the validity of the iosystem pointer before use.

  if (!(file = pio_get_file_from_id(ncid)))
￼          return PIO_EBADID;		￼         
￼      ios = file->iosystem;		￼   
￼ +    if(ios == NULL){
￼ +      fprintf(stderr, "%s:%d ERROR: File has no iosystem attached to it\n", __FILE__, __LINE__);
￼ +      return PIO_EBADID;
￼ +    }

This change was made in many places. Instead, I have modified the pio_get_file_from_id2() function, to check for this condition. (I also renamed it pio_get_file()). Then I changed the code to use that function, and return any error code.

So now this check (for a valid iosystem) is run within the function. A PIO_EINVAL is returned if the iosystem is NULL, which it should never be.

Since this pull request touches a lot of code I thought I would do the rest of the parameter checking changes in another pull request.

### DIFF
--- a/src/clib/pio_internal.h
+++ b/src/clib/pio_internal.h
@@ -78,7 +78,6 @@ extern "C" {
     int pio_delete_iodesc_from_list(int ioid);
     int pio_num_iosystem(int *niosysid);
 
-    file_desc_t *pio_get_file_from_id(int ncid);
     int pio_get_file(int ncid, file_desc_t **filep);
     int pio_delete_file_from_list(int ncid);
     void pio_add_to_file_list(file_desc_t *file);
--- a/src/clib/pio_internal.h
+++ b/src/clib/pio_internal.h
@@ -79,7 +79,7 @@ extern "C" {
     int pio_num_iosystem(int *niosysid);
 
     file_desc_t *pio_get_file_from_id(int ncid);
-    int pio_get_file_from_id2(int ncid, file_desc_t **filep);
+    int pio_get_file(int ncid, file_desc_t **filep);
     int pio_delete_file_from_list(int ncid);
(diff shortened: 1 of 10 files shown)


src/clib/pio_internal.h
```diff
@@ -79,7 +79,7 @@ extern "C" {
     int pio_num_iosystem(int *niosysid);
 
     file_desc_t *pio_get_file_from_id(int ncid);
-    int pio_get_file_from_id2(int ncid, file_desc_t **filep);
+    int pio_get_file(int ncid, file_desc_t **filep);
```
jayeshkrishna: Why not just replace pio_get_file_from_id() with this new interface (instead of adding a new function - pio_get_file() )?

src/clib/pio_lists.c
```diff
@@ -41,27 +41,26 @@ void pio_add_to_file_list(file_desc_t *file)
...
-	for (cfile = pio_file_list; cfile; cfile = cfile->next)
-	    if (cfile->pio_ncid == ncid)
-	    {
-		current_file = cfile;
-		break;
-	    }
-
-    return cfile;
-}
+/* file_desc_t *pio_get_file_from_id(int ncid) */
```
jayeshkrishna: Remove this function?

src/clib/pio_internal.h
edhartnett: I did replace it. Just didn't take the prototype out. But now I have.

So the end result is there is only one function to do this, pio_get_file().

PR #129, opened 2016-10-20
reindented, untabified, and deleted trailing whitespace

edhartnett:
A good suggestion from the meeting is that reformatting be applied to the entire codebase, without any other changes, as its own pull request.

That is what this is. In emacs I opened each file and ran the following commands:
M-x indent-region (for the whole file)
M-x untabify
M-x delete-trailing-whitespace

So there are no actual code changes. Just a lot of whitespace changes.

### DIFF
--- a/src/clib/bget.c
+++ b/src/clib/bget.c
@@ -1,401 +1,401 @@
 /*
 
-			       B G E T
-
-			   Buffer allocator
-
-    Designed and implemented in April of 1972 by John Walker, based on the
-    Case Algol OPRO$ algorithm implemented in 1966.
-
-    Reimplemented in 1975 by John Walker for the Interdata 70.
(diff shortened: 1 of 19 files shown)


rljacob:
Before merging this, we should make sure no one has any branches in progress that they are ready to merge.  This needs to be a sync point in development.

rljacob:
No branches that are touching the same files, that is.

edhartnett:
@rljacob do you have such a branch out?

How about you @jayeshkrishna?

jayeshkrishna:
I don't think I have any branches that touches these sources. PR #132 should have all the changes (only touches the fortran tests and the testing framework - tests/general/*) I need to get merged now.

rljacob:
I also don't have any branches outstanding.

PR #146, opened 2016-10-24
This enables internal and external documentation

### LINKED ISSUE #145: Produce internal (developer) documentation build
We have been using the @internal flag to mark functions as internal. It would be great if it did that, but it doesn't. ;-)

If we want to have internal vs. user documentation (and I believe we do) then we need to find another way.

I think the easiest way is going to be to designate some files as public, and have all the public (and no internal) functions in them. Then have all the internal functions in files with no public functions. If that was the case, it would be very easy to build the subset of public files for the user documentation, and all files for the developer documentation.

edhartnett:
The user facing documentation should not have our internal functions. 

But we do want to have developer documentation of our internal functions as well.

So with this PR I introduce the PIO_INTERNAL_DOC flag to CMake (default = Off). When on, all function documentation is packaged up. When not on, only a subset of files in the C api are used, the ones that contain public functions.

Some functions will have to be moved around in a future PR, in order to get all the functions in the correct documentation build.

Fixes #145


edhartnett:
Good question! 

For now I think it should be the developer docs.

Katetc:
I tried to send this response as an email this morning and it didn't go through. Posting it now for posterity. 

I wonder how hard it would be to reorganize the website site to support both? We could have a section describing the "PIO User Interface" and another with "PIO Developer API". Or, just a special, highlighted section for the user interface and then the regular functional documentation around as usual (which is similar to how we have it now). I could look into this once the functions have been re-tagged and reorganized. I don't think the website should be one or the other, IMHO.

edhartnett:
I agree that both forms of the documentation should be available. And most people should be directed to the user documentation, not the developer documentation. 

I believe we need to have a PIO web landing page for users.

It will have:
* links to current and past release tarballs.
* links to current and past documentation for users and developers.

I have not really checked out how well github can meet these needs. 

Do projects ever use github as their sole landing page? Mostly I have seen projects use a web landing page for users, and github for developers.

Katetc:
I wonder how hard it would be to reorganize the website site to support
both? We could have a section describing the "PIO User Interface" and
another with "PIO Developer API". Or, just a special, highlighted section
for the user interface and then the regular functional documentation around
as usual (which is similar to how we have it now). I could look into this
once the functions have been re-tagged and reorganized. I don't think the
website should be one or the other, IMHO.

PR #154, opened 2016-10-26
documentation fixes

### LINKED ISSUE #152: Changes to doxygen configure broke CDash runs
Looks like my recent changes to the doxygen files causes the configure to break on yellowstone.

Really, the docs config failing should not break the build, just disable the docs.

edhartnett:
This PR contains some doxygen changes.

It also adds the option of not building the documentation, and this option is used on yellowstone ctest builds. 

This should allow cdash tests to pass again.

We should allow builds without the documentation build. The documentation will be available on line, and not every target platform will have a recent version of Doxygen installed.

Fixes #152


CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required (VERSION 2.8.12)
 project (PIO C Fortran)
-cmake_policy(VERSION 3.5.2)
+#cmake_policy(VERSION 3.5.2)
```
jedwards4b: Why is this removed?

ctest/CTestEnvironment-nwsc.cmake
```diff
@@ -10,7 +10,7 @@
 # set with existing environment variables: NETCDF, PNETCDF, HDF5, etc.
 
 # Define the extra CMake configure options
-set (CTEST_CONFIGURE_OPTIONS "-DCMAKE_VERBOSE_MAKEFILE=TRUE -DPIO_ENABLE_ASYNC=TRUE")
+set (CTEST_CONFIGURE_OPTIONS "-DCMAKE_VERBOSE_MAKEFILE=TRUE -DPIO_ENABLE_DOC=OFF")
```
jedwards4b: Dont you still need -DPIO_ENABLE_ASYNC=TRUE?

edhartnett: No, that option went away with the full merger of the async code.

CMakeLists.txt
jayeshkrishna: Why do we want to have cmake_policy version higher than the min cmake version required to build PIO?

edhartnett: @jayeshkrishna that is a good question. Actually I had just set it to the version of doxygen that I have installed on my local machine.

PR #156, opened 2016-10-27
Adding a test with unlimited dim

jayeshkrishna:
This test writes out a 4d (one unlimited dim) array and reads
it back. Each frame is a 3d col decomposed data.


jayeshkrishna:
This test should pass.

PR #161, opened 2016-10-28
Adding a 1d decomp test with holes

jayeshkrishna:
This commit,
* Moves 1d decomp test to a separate test/file, pio_decomp_tests_1d.F90.in (leaving generic decomp tests in pio_decomp_tests.F90.in)
* Adds a new 1d decomp test with holes (some processes don't write any data)


jayeshkrishna:
These tests should all pass with the latest version of netcdf libraries (c-4.4.1-cxx-4.2-fortran-4.4.4).
The pio_decomp_tests_1d could fail with earlier versions of netcdf (failed for me with c-4.3.3.1-cxx-4.2-fortran-4.4.2 on my laptop).

edhartnett:
@jayeshkrishna I am very curious as to what failed in earlier versions of netCDF, but passes in 4.4.1.

Also surely the C++ API is not involved, is it? You guys are writing fortran so you would be using the netCDF fortran anc C libraries, right?

If this test fails for earlier versions of netCDF does that mean that we require netCDF-c 4.4.1 and netCDF-fortran 4.4.4? If so, we should build that into the configure step, so that the user gets a clear error message and can upgrade netCDF (which I suspect can be very easily done).

jayeshkrishna:
Yes, we need to use the latest versions c-4.4.1 + cxx-4.2 + fortran-4.4.4 if we plan to use NETCDF4P. Otherwise the older versions should be good.
I have added some information here - https://github.com/NCAR/ParallelIO/pull/162 . Its probably a bug in netcdf when some processes don't write (count[*] = 0) data.

PR #162, opened 2016-10-28
Adding BOX rearranger tests

jayeshkrishna:
Adding code to test each case with the BOX rearranger (and the
SUBSET rearranger). We were only testing the SUBSET rearranger
prior to this change


jayeshkrishna:
These tests should all pass with the latest version of netcdf libraries (c-4.4.1-cxx-4.2-fortran-4.4.4). The tests failed for the BOX rearranger with NETCDF4P with an older version of the netcdf libraries (c-4.3.3.1-cxx-4.2-fortran-4.4.2) on my laptop.
Googling around this problem is the same (/similar) as the one reported a while back to netcdf - see https://www.unidata.ucar.edu/support/help/MailArchives/netcdf/msg11308.html (netCDF #PDL-125161)

PR #166, opened 2016-10-28
fixed naughty behavior with finalize

### LINKED ISSUE #163: Some unit tests fail with the same MPI error
When PIO is built with GCOV flags, 9 unit tests fail with the same MPI error.
4:test_async_simple
5:test_async_3proc
6:test_async_4proc
7:test_iosystem2_simple
8:test_iosystem2_simple2
9:test_iosystem2
10:test_iosystem3_simple
11:test_iosystem3_simple2
12:test_iosystem3

test_async_simple fails like this:
Error 1111 in tests/unit/test_async_simple.c, line 110
Attempting to use an MPI routine after finalizing MPICH

edhartnett:
As pointed out in github issue #163 some of the C tests were calling MPI_Finalize() twice.

This PR fixes the bugs that caused that, including lack of return statement, and calling ERR() macro on failure of pio_test_finalize() (which will have already called MPI_finalize()).

Fixes #163.

### DIFF
--- a/tests/unit/test_async_2comp.c
+++ b/tests/unit/test_async_2comp.c
@@ -124,7 +124,7 @@ main(int argc, char **argv)
     /* Finalize the MPI library. */
     printf("%d %s Finalizing...\n", my_rank, TEST_NAME);
     if ((ret = pio_test_finalize()))
-	ERR(ERR_AWFUL);
+	return ERR_AWFUL;
 
     printf("%d %s SUCCESS!!\n", my_rank, TEST_NAME);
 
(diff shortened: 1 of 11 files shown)


dqwu:
Looks good. For a few files like test_iosystem3.c, a lot of line changed are just related to formatting? This has caused the file diffs to be very long.

edhartnett:
Yes, looks like someone re-indented it between the time that I originally checked it in and the time we agreed to all use the same indents. ;-) So I reindented it to our project standard.

edhartnett:
Jim do you want me to merge this PR?

rljacob:
Formatting changes should always be done as a separate PR.  Even when its fixing a mistake.

Issue #120, opened 2016-10-17
Implement PIO1 rearranger flow control options in PIO2


jayeshkrishna:
PIO1 allows user to set flow control options (the internal flow
control parameters that determine how the data exchange
during data rearrangement works) at runtime by passing
the information via PIO_Init. We need to implement this
in PIO2.
The related Fortran namelists were added to PIO2 in
6aa3586fd2077aca74f39471488abcad17555b1a . See
PR #115

jedwards4b:
What is the status of this work?

jayeshkrishna:
Not started yet. Will start soon.

jayeshkrishna:
This work is in progress.

jayeshkrishna:
PR #479 fixed this issue.

PR #171, opened 2016-11-01
removed use of environmental vars

### LINKED ISSUE #138: fix warnings in pioc_support.c
There are some warnings that should be fixed:

/home/ed/pio3/ParallelIO/src/clib/pioc_support.c: In function ‘pio_get_env’:
/home/ed/pio3/ParallelIO/src/clib/pioc_support.c:195:26: warning: passing argument 2 of ‘strchr’ makes integer from pointer without a cast [-Wint-conversion]
         if(strchr(envptr,"M") != NULL){
                          ^~~
In file included from /home/ed/pio3/ParallelIO/src/clib/pio.h:15:0,
                 from /home/ed/pio3/ParallelIO/src/clib/pioc_support.c:9:
/usr/include/string.h:231:14: note: expected ‘int’ but argument is of type ‘char *’
 extern char *strchr (const char *__s, int __c)
              ^~~~~~
/home/ed/pio3/ParallelIO/src/clib/pioc_support.c:197:32: warning: passing argument 2 of ‘strchr’ makes integer from pointer without a cast [-Wint-conversion]
         }else if(strchr(envptr,"K") != NULL){
                                ^~~
In file included from /home/ed/pio3/ParallelIO/src/clib/pio.h:15:0,
                 from /home/ed/pio3/ParallelIO/src/clib/pioc_support.c:9:
/usr/include/string.h:231:14: note: expected ‘int’ but argument is of type ‘char *’
 extern char *strchr (const char *__s, int __c)
              ^~~~~~
/home/ed/pio3/ParallelIO/src/clib/pioc_support.c: In function ‘PIOc_writemap_from_f90’:
/home/ed/pio3/ParallelIO/src/clib/pioc_support.c:649:54: warning: passing argument 5 of ‘PIOc_writemap’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
     return(PIOc_writemap(file, ndims, gdims, maplen, map, MPI_Comm_f2c(f90_comm)));
                                                      ^~~
/home/ed/pio3/ParallelIO/src/clib/pioc_support.c:581:5: note: expected ‘MPI_Offset \* {aka long long int *}’ but argument is of type ‘const long long int *’
 int PIOc_writemap(const char file[], const int ndims, const int gdims[], PIO_Offset maplen, PIO_Offset map[], const MPI_Comm comm)
     ^~~~~~~~~~~~~

### LINKED ISSUE #143: Clarify requirements wrt environment variables
In pioc_support.c, there is a function:

/\* Get PIO environment variables.
 */
void pio_get_env(void)

It is called only once, in pioc.c at the end of Init_Intracomm().

As discussed at the recent meeting, we don't want to use environment variables moving forward with netCDF. So the questions are: 
1. What settings are configurable by environment variables.
2. Do we need to support some environment variables for backward compatibility?

edhartnett:
From discussion with Jim, it seemed that the use of environment variables in Init_Intracomm() was an old feature, no longer used.

Environment variables are not a good way to control our library functions, so that is probably why their use was abandoned. 

There was never any reference to the environment vars in Init_Intercomm().

This pull request removes this old code.

Fixes #143 
Fixes #138

### DIFF
--- a/src/clib/pioc.c
+++ b/src/clib/pioc.c
@@ -528,9 +528,6 @@ int PIOc_Init_Intracomm(const MPI_Comm comp_comm, const int num_iotasks,
         /* Add this iosys struct to the list in the PIO library. */
         *iosysidp = pio_add_to_iosystem_list(iosys);
 
-        /* Get some info from the environment. */
-        pio_get_env();
-
         /* Allocate buffer space for compute nodes. */
         compute_buffer_init(*iosys);
 
--- a/src/clib/pioc_support.c
+++ b/src/clib/pioc_support.c
@@ -177,51 +177,6 @@ void pio_log(int severity, const char *fmt, ...)
 static pio_swapm_defaults swapm_defaults;
 bool PIO_Save_Decomps=false;
 
-/** Get PIO environment variables.
- */
-void pio_get_env(void)
-{
-    char *envptr;
-    extern bufsize pio_cnbuffer_limit;
-    envptr = getenv("PIO_Save_Decomps");
-
-    if(envptr != NULL && (strcmp(envptr,"true")==0)){
-        PIO_Save_Decomps=true;
-    }
-    swapm_defaults.nreqs = 0;
-    swapm_defaults.handshake=false;
-    swapm_defaults.isend=false;
-
-    envptr = getenv("PIO_SWAPM");
-    if(envptr != NULL){
-        char *token = strtok(envptr, ":");
-
-        swapm_defaults.nreqs = atoi(token);
-
-        token = strtok(NULL, ":");
-
-        if((token!=NULL) && strcmp(token,"t")==0){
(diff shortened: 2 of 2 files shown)


Katetc:
Actually, I have used these environment variables in testing PIO2 performance. Is there another way to set them besides the environment variables?

edhartnett:
I can add another way to set them, if it does not already exist in the API.

What are you setting?

Katetc:
I set handshake to true and nreqs to 64 to mimic the settings in PIO1 and see if that helped performance. It did a little but not enough. Anyway, it would be nice to keep a way to set those around, because we may use them in CESM.

jedwards4b:
Issue #120 includes the list of settings from pio1 that we also want to provide control for in pio2.  
The pio_save_decomps and pio_cnbuffer_limit should also have api interfaces.

Issue #89, opened 2016-08-04
Implement async versions of darray functions


edhartnett:

edhartnett:
These are now working.

PR #186, opened 2016-11-10
Comments for PIOc_write_darray()

edhartnett:
I have added comments to PIOc_write_darray(). Please review the comments and tell me what I got wrong. ;-)

I have used double-question marks where I am confused. If you could provide some comments for those places, that would really speed things up.

This is related to #89.

### DIFF
--- a/src/clib/pio_darray.c
+++ b/src/clib/pio_darray.c
@@ -283,8 +283,8 @@ int PIOc_write_darray(const int ncid, const int vid, const int ioid,
     /* Is this a record variable? */
     recordvar = vdesc->record >= 0 ? true : false;
 
-    /* Check that ??? is not the same as length of data being
-     * written? */
+    /* Check that the local size of the variable passed in matches the
+     * size expected by the io descriptor. */
     if (iodesc->ndof != arraylen)
         piodie("ndof != arraylen",__FILE__,__LINE__);
 
@@ -305,7 +305,10 @@ int PIOc_write_darray(const int ncid, const int vid, const int ioid,
         /* separate record and non-record variables */
         if (recordvar)
         {
-	    /* What is going on here?? */
+	    /* wmb is write multi buffer and is a pointer to one or
+	     * more variables that fit the same description. We are
+	     * moving to the end of the wmb linked list to add the
+	     * current variable. */
             while(wmb->next && wmb->ioid != ioid)
                 if (wmb->next)
                     wmb = wmb->next;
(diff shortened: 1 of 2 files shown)


src/clib/pio_darray.c
```diff
@@ -361,34 +382,38 @@ int PIOc_write_darray(const int ncid, const int vid, const int ioid,
...
     if (arraylen > 0)
-        if (!(wmb->data = bgetr(wmb->data, (1+wmb->validvars)*arraylen*tsize)))
+        if (!(wmb->data = bgetr(wmb->data, (1 + wmb->validvars) * arraylen * tsize)))
             piomemerror(*ios, (1 + wmb->validvars) * arraylen * tsize, __FILE__, __LINE__);
 
     /* Get memory for ??? */
     if (!(wmb->vid = (int *)bgetr(wmb->vid, sizeof(int) * (1 + wmb->validvars))))
         piomemerror(*ios, (1 + wmb->validvars) * sizeof(int), __FILE__, __LINE__);
 
+    /* Get memory for ??? */
```
jedwards4b: wmb->frame is the record number, we assume that the variables in the wmb list may not all have the same unlimited dimension value although they usually do.

src/clib/pio_darray.c
```diff
@@ -407,32 +432,27 @@ int PIOc_write_darray(const int ncid, const int vid, const int ioid,
...
-      double asum=0.0;
-      printf("%s %d %d %d %d\n",__FILE__,__LINE__,vid,arraylen,iodesc->ndof);
-      for (int k=0;k<arraylen;k++){
-      asum += ((double *) array)[k];
-      }
-      printf("%s %d %d %g\n",__FILE__,__LINE__,vid,asum);
-      }
-    */
 
+    /* ??? */
```
jedwards4b: add the unlimited dimension value of this variable to the frame array in wmb

tests/unit/test_darray.c
```diff
@@ -1,7 +1,5 @@
-/**
- * @file
+/*
```
jedwards4b: You've tried to simplify this test and that's fine, but instead of replacing a working test why not create a new one?

edhartnett: This is not really a working test. If you look, you will see it does nothing with darrays (yet). It is just the start of the darray test I am writing. (The same is true of test_darray_async.c) They started as code copied from a test creating a more complex test file. In this first darray outing I am going to keep it very simple. ;-)

PR #203, opened 2016-11-16
need some comments in pio.h

edhartnett:
There are numerous "???" in the comments for the elements of structs io_region, io_desc_t, and wmulti_buffer.

In pio_darray.c there are also some "???" where I don't know what comment to put.

Can you use this review to tell me what these comments should be?

### DIFF
--- a/src/clib/pio_darray.c
+++ b/src/clib/pio_darray.c
@@ -73,9 +73,9 @@ int PIOc_write_darray_multi(const int ncid, const int *vid, const int ioid,
     iosystem_desc_t *ios;  /* Pointer to io system information. */
     file_desc_t *file;     /* Pointer to file information. */
     io_desc_t *iodesc;     /* Pointer to IO description information. */
-    int vsize;             /* ??? */
-    int rlen;              /* ??? */
-    var_desc_t *vdesc0;    /* ??? */
+    int vsize;             /* size in bytes of the given data. */
+    int rlen;              /* total data buffer size. */
+    var_desc_t *vdesc0;    /* pointer to var_desc structure for each var. */
     int ierr = PIO_NOERR;  /* Return code. */
 
     /* Check inputs. */
@@ -95,7 +95,11 @@ int PIOc_write_darray_multi(const int ncid, const int *vid, const int ioid,
     if (!(iodesc = pio_get_iodesc_from_id(ioid)))
         return PIO_EBADID;
 
-    /* ??? */
+    /* For netcdf serial writes we collect the data on io nodes and
+     * then move that data one node at a time to the io master node
+     * and write (or read). The buffer size on io task 0 must be as
+     * large as the largest used to accommodate this serial io
+     * method. */
     vdesc0 = file->varlist + vid[0];
 
     /* ??? */
@@ -107,7 +111,8 @@ int PIOc_write_darray_multi(const int ncid, const int *vid, const int ioid,
     if (vdesc0->iobuf)
         piodie("Attempt to overwrite existing io buffer",__FILE__,__LINE__);
 
-    /* ??? */
+    /* Currently there are two rearrangers box=1 and subset=2. There
+     * is never a case where rearranger==0. */
     if (iodesc->rearranger > 0)
     {
         if (rlen > 0)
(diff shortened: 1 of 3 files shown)


src/clib/pio_darray.c
```diff
@@ -68,11 +71,12 @@ int PIOc_write_darray_multi(const int ncid, const int *vid, const int ioid,
...
-    file_desc_t *file;
-    io_desc_t *iodesc;
-    int vsize, rlen;
-    int ierr = PIO_NOERR;
-    var_desc_t *vdesc0;
+    file_desc_t *file;     /* Pointer to file information. */
+    io_desc_t *iodesc;     /* Pointer to IO description information. */
+    int vsize;             /* ??? */
+    int rlen;              /* ??? */
+    var_desc_t *vdesc0;    /* ??? */
```
jedwards4b: *vid is a list of variable Ids to be written and vdesc is a pointer to the var_desc structure associated with each variable.

jedwards4b: vsize is the unit size in bytes of the given data.  rlen is the total data buffer size.

src/clib/pio_darray.c
```diff
@@ -91,16 +95,19 @@ int PIOc_write_darray_multi(const int ncid, const int *vid, const int ioid,
...
+    /*   rlen = iodesc->llen*nvars; */
     rlen=0;
     if (iodesc->llen > 0)
         rlen = iodesc->maxiobuflen*nvars;
 
     if (vdesc0->iobuf)
         piodie("Attempt to overwrite existing io buffer",__FILE__,__LINE__);
 
+    /* ??? */
     if (iodesc->rearranger > 0)
```
jedwards4b: I originally conceived of an option in which the data could be provided to pio already in an arrangement suitable for passing on to the underlying io layer, but I've yet to find a case where that is true and so rearranger is always > 0.   Currently there are two rearrangers box=1 and subset=2.   There is never a case where rearranger==0

src/clib/pio_darray.c
```diff
@@ -162,6 +170,7 @@ int PIOc_write_darray_multi(const int ncid, const int *vid, const int ioid,
         }
     }
 
+    /* ??? */
```
jedwards4b: I have separately shared a set of slides with you that illustrate the two different rearrangement options provided.   The box rearranger will always have data (it could be fill data) to fill the entire array - that is the aggregate start and count values will completely describe one unlimited dimension unit of the array.  For the subset method this is not necessarily the case, areas of missing data may never be written.  In order to make sure that these areas are given the missing value a 'holegrid' is used to describe the missing points.   This is generally faster than the netcdf method of filling the entire array with missing values before overwriting those values later.

src/clib/pio_darray.c
```diff
@@ -171,6 +180,7 @@ int PIOc_write_darray_multi(const int ncid, const int *vid, const int ioid,
         /* Get a buffer. */
         vdesc0->fillbuf = bget(iodesc->holegridsize * vsize * nvars);
 
+	/* ??? */
```
jedwards4b: copying the fill value into the data buffer for the box rearranger.  This will be overwritten with data where provided.

src/clib/pio_darray.c
```diff
@@ -323,13 +333,14 @@ int PIOc_write_darray(const int ncid, const int vid, const int ioid,
...
         else
         {
+	    /* Move to end of list. */
             while(wmb->next && wmb->ioid != -(ioid))
                 if (wmb->next)
                     wmb = wmb->next;
         }
     }
 
-    /* ?? */
```
jedwards4b: the write multi buffer wmulti_buffer is the cache on compute nodes that will collect and store multiple variables 
before sending them to the io nodes.   Aggregating variables in this way leads to a considerable savings in communication expense.  Variables in the wmb array must have the same decomposition and base data size and we also need to keep track of whether each is a recordvar (has an unlimited dimension) or not.

src/clib/pio_darray.c
```diff
@@ -482,21 +493,20 @@ int PIOc_write_darray(const int ncid, const int vid, const int ioid,
...
 {
     iosystem_desc_t *ios;  /* Pointer to io system information. */
-    file_desc_t *file;
-    io_desc_t *iodesc;
-    void *iobuf = NULL;
-    size_t rlen = 0;
-    int tsize; /* Total size. */
+    file_desc_t *file;     /* Pointer to file information. */
+    io_desc_t *iodesc;     /* Pointer to IO description information. */
+    void *iobuf = NULL;    /* ??? */
```
jedwards4b: iobuf holds the data as read on the io node. 
rlen is the length of that buffer.

src/clib/pio_darray.c
```diff
@@ -507,6 +517,7 @@ int PIOc_read_darray(const int ncid, const int vid, const int ioid,
     if (!(iodesc = pio_get_iodesc_from_id(ioid)))
         return PIO_EBADID;
 
+    /* ??? */
```
jedwards4b: For netcdf serial writes we collect the data on io nodes and then move that data one node at a time to the io master node and write (or read).   The buffer size on io task 0 must be as large as the largest used to accommodate this serial io method.

src/clib/pio_darray.c
edhartnett: As a performance note, netCDF-4 files do not write fill data. If you don't write a portion of a variable it is not written in the file. If you later try to read it, fill values are provided. But the fill values don't get written.

So it seems like for netCDF-4 you should never try to add fill values.

edhartnett:
I have added all your comments to the branch.

The comments are super-helpful and I already have a much better understanding of what is going on.

Can you take a look at the many ??? in pio.h?

PR #209, opened 2016-11-17
Changes to support async tests on Laramie - needed different indexes …

Katetc:
…and tags in the create_intercomm call in init. @edhartnett Please take a look at these changes. The Laramie test cluster wanted some changes to the MPI_Intercomm_create call to get it to work. I changed the index of the second comm root to be relative to the peer comm (union_comm) and the tag to be a single value (rather than all different).


Katetc:
Eh, conflicts with my fork. I'll submit this from a branch later.

PR #215, opened 2016-11-21
Adding comments

edhartnett:
Just adding more comments based on continuing study of code and answers from Jim.

No code changes in this review.

Somehow we keep getting a lot of trailing whitespace. It might be my fault. I will see if I can sort that out and avoid it.

### DIFF
--- a/src/clib/pio_file.c
+++ b/src/clib/pio_file.c
@@ -117,7 +117,7 @@ int PIOc_createfile(const int iosysid, int *ncidp, int *iotype,
     file->buffer.data = NULL;
     file->buffer.next = NULL;
     file->buffer.vid = NULL;
-    file->buffer.ioid = 0;
+    file->buffer.ioid = -1;
     file->buffer.frame = NULL;
     file->buffer.fillvalue = NULL;
 
--- a/src/clib/pio_file.c
+++ b/src/clib/pio_file.c
@@ -117,7 +117,7 @@ int PIOc_createfile(const int iosysid, int *ncidp, int *iotype,
     file->buffer.data = NULL;
     file->buffer.next = NULL;
     file->buffer.vid = NULL;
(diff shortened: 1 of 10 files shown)


src/clib/pio_file.c
```diff
@@ -117,7 +117,7 @@ int PIOc_createfile(const int iosysid, int *ncidp, int *iotype,
     file->buffer.data = NULL;
     file->buffer.next = NULL;
     file->buffer.vid = NULL;
-    file->buffer.ioid = -1;
+    file->buffer.ioid = 0;
```
jedwards4b: I think that there was a reason that this was -1 and not 0 - have you tested this change?

src/clib/pio_rearrange.c
```diff
@@ -738,6 +746,7 @@ int rearrange_comp2io(const iosystem_desc_t ios, io_desc_t *iodesc, void *sbuf,
         sendtypes[i] =  MPI_DATATYPE_NULL;
     }
 
+    /* ??? */
```
jedwards4b: If this io proc will exchange data with compute tasks create a MPI DataType for that exchange.

src/clib/pio_rearrange.c
```diff
@@ -776,6 +785,7 @@ int rearrange_comp2io(const iosystem_desc_t ios, io_desc_t *iodesc, void *sbuf,
         }
     }
 
+    /* ??? */
```
jedwards4b: On compute tasks loop over iotasks and create a data type for each exchange.

src/clib/pio_file.c
edhartnett: I didn't intend for this to be part of the PR. I was indeed testing it out, but decided that I wanted to stick to comment so real code changes could be reviewed more carefully.

PR #219, opened 2016-11-22
Now all C tests can handle extra tasks.

### LINKED ISSUE #211: C tests can't really handle more than 4 tasks...
The C tests are supposed to be written so that more than the specified number of tasks can be used.

But in fact, when I run a 4-task test with 6 processors, it hangs at the end of the test and does not return.

edhartnett:
Fixes #211.

There are only test code changes in this PR - no library code changes.

Many tests could already handle extra tasks, but some could not, because they were calling InitIntracomm with MPI_COMM_WORLD (instead of test_comm), and/or using ntasks for calculations, instead of TARGET_NTASKS.

Those don't matter if they run on exactly the correct number of tasks, but they cause failures if extra tasks are included.

To test all C tests are now run with an extra task.

Also removes doxygen tags from test code (since we don't run doxygen on tests).

### DIFF
--- a/tests/unit/test_intercomm2.c
+++ b/tests/unit/test_intercomm2.c
@@ -7,11 +7,8 @@
  * --error-exitcode=99 --track-origins=yes ./test_intercomm2</pre>
  *
  */
-#define LOGGING 1
 #include <pio.h>
 #include <pio_tests.h>
-#include <netcdf.h>
-#include <unistd.h>
 
 /* The number of tasks this test should run on. */
 #define TARGET_NTASKS 4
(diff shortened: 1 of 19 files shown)


tests/unit/test_intercomm2.c
```diff
@@ -212,8 +198,7 @@ check_file(int iosysid, int format, char *filename, int my_rank, int verbose)
         ERR(ERR_WRONG);
     if ((ret = PIOc_get_att_int(ncid, NC_GLOBAL, ATT_NAME, &att_data)))
         ERR(ret);
-    if (verbose)
-        printf("%d test_intercomm2 att_data = %d\n", my_rank, att_data);
+    printf("%d test_intercomm2 att_data = %d\n", my_rank, att_data);
```
jedwards4b: why did you remove the verbose flag?

edhartnett: Because the build system already provides the feature that tests will run without output. I expected originally that my test print statements would pollute the output with masses of lines of information, so I would have to turn them off with a verbose flag.

But since I don't need to do that, it was always set to 1 and was therefore just useless code.

Issue #205, opened 2016-11-16
Bring test coverage of C library to maximum possible


edhartnett:
Thanks to @dqwu we have some code coverage data. For the C library the overall 64% of lines covered.

I think a goal of > %90 is achievable and desirable. 

Adding tests is easy, if somewhat tedious. 

Worst offenders at the moment are pioc_get/put_nc.c, pio_spmd.c, and pio_nc4.c.

edhartnett:
We are up to 69% code coverage for the C library.

I just went through each code file and added a ticket for each function that was untested.

edhartnett:
As of today we are up to 78% code coverage, which is a nice improvement!

edhartnett:
As of today we are at %80.

edhartnett:
We are now at ~84% coverage.

It's going to be hard to get much higher. Almost all of the uncovered lines are handling of MPI or other failure that we cannot easily make happen in a test.

So I will change the name of this issue to say "maximum" instead of "90%".

edhartnett:
Current coverage is:
```
Lines:	5782	7051	82.0 %
Functions:  	355	369	96.2 %
Branches:	4230	6041	70.0 %

```
I am going to close this ticket. I will continue to monitor testing as new code is added, but I believe we've reached pretty good test coverage for existing code, which was what this ticket was about.

PR #221, opened 2016-11-22
fixed bug in PIOc_open()

edhartnett:
Seeking to add to the test coverage, I found a bug in PIOc_open().

Unlike PIOc_openfile(), PIOc_open() is a copy of the nc_open() function prototype. It passes in a mode parameter, instead of an iotype. It was added for netCDF integration and is not used by any existing PIO users.

The mode is more versatile, because there are more available modes than are expressed in iotypes currently. It also irons out a major issue in the upcoming netCDF integration. (There is no intention of taking PIOc_openfile() away.)

However there was a bug in PIOc_open() in the way I was determining the iotype for netCDF-4 parallel cases, which resulted in using netCDF-4 serial instead.

So on the very first line I tried to increase test coverage, I found a bug. What a great reminder how important it is to have good test coverage! ;-)

This is in aid of #205.

### DIFF
--- a/src/clib/pio_file.c
+++ b/src/clib/pio_file.c
@@ -34,22 +34,24 @@ int PIOc_openfile(const int iosysid, int *ncidp, int *iotype,
     return PIOc_openfile_retry(iosysid, ncidp, iotype, filename, mode, 1);
 }
 
-/** Open an existing file using PIO library.
+/** 
+ * Open an existing file using PIO library.
  *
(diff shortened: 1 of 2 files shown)


jayeshkrishna:
@edhartnett : It looks like all the changes required in this PR was in the test program. I am not sure whether the changes made to the library code is related to fixing the issue noted in this PR.
IMO, We should separate out commits so that it is easy to identify what got fixed (and how).

PR #224, opened 2016-11-24
SPMD removing unused code, adding comments, reusing old test code

### LINKED ISSUE #207: Remove main() from pio_spmd.c
In pio_spmd.c there is a main() with the following comment:

/*
  This program tests MPI_Alltoallw by having processor i send different
  amounts of data to each processor.
  The first test sends i items to processor i from all processors.
*/

Judging by other code, this is obsolete and should probably just be removed. If we do still want it, let's make it a test program in the test directory.

edhartnett:
Fixes #207.

In consultation with Jim, removed some unused code in pio_spmd. 

Also added comments, did some cleanup.

Also added the new test_spmd.c, still under development. This is a test that recovers the old test that was in pio_spmd.c main(). However the test does not yet work with current code. I will continue with the test under antoher github ticket.

Many "???" in this complex code. Jim, if you can contribute any comments, that would be very helpful.


src/clib/pio_spmd.c
```diff
@@ -305,108 +251,97 @@ int pio_swapm(void *sndbuf, int *sndlths, int *sdispls, MPI_Datatype *stypes,
...
           printf("%s %d %d %d\n",__FILE__,__LINE__,extent, lb);
-          MPI_Type_get_extent(rtypes[mytask], &lb, &extent);
+          MPI_Type_get_extent(recvtypes[my_rank], &lb, &extent);
           printf("%s %d %d %d\n",__FILE__,__LINE__,extent, lb);
         */
-#ifdef ONEWAY
-        CheckMPIReturn(MPI_Sendrecv(sptr, sndlths[mytask],stypes[mytask],
-                                    mytask, tag, rptr, rcvlths[mytask], rtypes[mytask],
-                                    mytask, tag, comm, &status),__FILE__,__LINE__);
+#ifdef ONEWAY /* ??? */
```
jedwards4b: ONEWAY or another - this is a bad name for this macro which if true will post mpi_sendrecv comms instead of irecv/send.

src/clib/pio_spmd.c
```diff
@@ -305,108 +251,97 @@ int pio_swapm(void *sndbuf, int *sndlths, int *sdispls, MPI_Datatype *stypes,
...
 #endif
     }
-    
-    if (nprocs == 1)
-        return PIO_NOERR;
 
-    int swapids[nprocs];
-    MPI_Request sndids[nprocs];
-    MPI_Request hs_rcvids[nprocs];
+    /* ??? */
```
jedwards4b: When send to self is complete there is nothing left to do if ntasks==1

src/clib/pio_spmd.c
```diff
@@ -438,32 +373,32 @@ int pio_swapm(void *sndbuf, int *sndlths, int *sdispls, MPI_Datatype *stypes,
...
                 CheckMPIReturn(MPI_Irecv(&hs, 1, MPI_INT, p, tag, comm, hs_rcvids + istep),
-			       __FILE__,__LINE__);
+                               __FILE__, __LINE__);
             }
         }
     }
-    
+
+    /* ??? */
     for (istep = 0; istep < maxreq; istep++)
```
jedwards4b: Post up to maxreq irecv's

src/clib/pio_spmd.c
```diff
@@ -438,32 +373,32 @@ int pio_swapm(void *sndbuf, int *sndlths, int *sdispls, MPI_Datatype *stypes,
...
+            ptr = (char *)recvbuf + rdispls[p];
 
-            CheckMPIReturn(MPI_Irecv(ptr, rcvlths[p], rtypes[p], p, tag, comm, rcvids+istep),
-			   __FILE__,__LINE__);
+            CheckMPIReturn(MPI_Irecv(ptr, recvcounts[p], recvtypes[p], p, tag, comm, rcvids + istep),
+                           __FILE__, __LINE__);
 
             if (handshake)
-                CheckMPIReturn(MPI_Send(&hs, 1, MPI_INT, p, tag, comm), __FILE__,__LINE__);
+                CheckMPIReturn(MPI_Send(&hs, 1, MPI_INT, p, tag, comm), __FILE__, __LINE__);
```
jedwards4b: Tell the paired task that this tasks' has posted it's irecvs'

src/clib/pio_spmd.c
```diff
@@ -472,50 +407,50 @@ int pio_swapm(void *sndbuf, int *sndlths, int *sdispls, MPI_Datatype *stypes,
...
-	{
-            tag = mytask + offset_t;
+        if (sendcounts[p] > 0)
+        {
+            tag = my_rank + offset_t;
             if (handshake)
-	    {
-                CheckMPIReturn(MPI_Wait(hs_rcvids+istep, &status), __FILE__,__LINE__);
+            {
+                CheckMPIReturn(MPI_Wait(hs_rcvids + istep, &status), __FILE__, __LINE__);
```
jedwards4b: If handshake is enabled don't post sends until the receiving task has posted recvs

src/clib/pio_spmd.c
```diff
@@ -525,210 +460,10 @@ int pio_swapm(void *sndbuf, int *sndlths, int *sdispls, MPI_Datatype *stypes,
     /* ??? */
     if (steps > 0)
```
jedwards4b: If steps > 0 there are still outstanding messages, wait for them here.

tests/unit/CMakeLists.txt
```diff
@@ -47,6 +47,9 @@ endif()
 
 # Don't run these tests if we are using MPI SERIAL.
 if (NOT PIO_USE_MPISERIAL)
+  add_executable (test_spmd EXCLUDE_FROM_ALL test_spmd.c test_common.c)
```
jedwards4b: Although these routines don't make a lot of sense with mpi serial they do work and should be tested.

tests/unit/test_spmd.c
```diff
@@ -0,0 +1,254 @@
...
+        if (!my_rank)
+        {
+            for (int e = 0; e < num_elem; e++)
+                printf("sbuf[%d] = %d", e, sbuf[e]);
+            for (int e = 0; e < num_elem; e++)
+                printf("rbuf[%d] = %d", e, rbuf[e]);
+        }
+    }
+
+    /* Test pio_fc_gather. In fact it does not work for msg_cnt > 0. */
```
jedwards4b: Should test pio_fc_gatherv

tests/unit/test_spmd.c
```diff
@@ -0,0 +1,254 @@
...
+        MPI_Barrier(test_comm);
+
+        if (!my_rank)
+        {
+            printf("Start itest %d\n", itest);
+            gettimeofday(&t1, NULL);
+        }
+
+        if (itest == 0)
+            ret = pio_swapm(sbuf, sendcounts, sdispls, sendtypes, rbuf, recvcounts,
```
jedwards4b: we really aren't testing much if the only test is with msg_cnt==0

src/clib/pio_spmd.c
```diff
@@ -305,108 +251,97 @@ int pio_swapm(void *sndbuf, int *sndlths, int *sdispls, MPI_Datatype *stypes,
...
+    /* Send to self. */
+    if (sendcounts[my_rank] > 0)
     {
         void *sptr, *rptr;
         int extent, lb;
-        tag = mytask + offset_t;
-        sptr = (void *)((char *) sndbuf + sdispls[mytask]);
-        rptr = (void *)((char *) rcvbuf + rdispls[mytask]);
+        tag = my_rank + offset_t;
+        sptr = (char *)sendbuf + sdispls[my_rank];
```
edhartnett: The char was already there (and is needed to make the pointer arithmetic work). The cast of the result to void * is not necessary. It will happen anyway when sptr (which is a void *) gets the char * which is the result of the expression. It's part of the definition of void - we don't need to do it in our code.

tests/unit/test_spmd.c
edhartnett: I agree! I wanted to get these changes merged, then will move on with this test.

Now that I have added your very helpful comments I will do the merge. Thanks!! ;-)

tests/unit/test_spmd.c
edhartnett: Yes, it will be. This is just the commented out code of my test for pio_fc_gather(), which turned out not to be needed, but looks very similar to what will be needed for pio_fc_gatherv(). After I get the swapm test working, I will uncomment this code and turn it into a test for pio_fc_gatherv().

Issue #225, opened 2016-11-24
Resolve inconsistent error handling in C library


edhartnett:
The C code responds to errors in a variety of ways, and we should settle on the best way, and use it everywhere.

Some of the ways currently in use:
* Ignore return codes - for some calls we ignore the result.
* Return error code - good in normal C library, but sometimes causes one or more tasks to hang and program has to time out.
* Print error message and continue - error messages are helpful and need to be preserved, but we should not continue after error.

Our requirements:
* Code must stop and return error as soon as an error is detected.
* Sensible error messages are provided to users.
* PIO error handler for file is followed, if error occurs on file access.
* PIO default error handler is followed, if error occurs before file is open. (?)
* We must handle 3 kinds of errors: netCDF, MPI, and PIO.

What should we do?

Here's an example for MPI Calls:
```
    int mpierr;  /* Return code from MPI functions. */    

        if ((mpierr = MPI_Alltoallw(sendbuf, sendcounts, sdispls, sendtypes, recvbuf,
                                    recvcounts, rdispls, recvtypes, comm)))
            return check_mpi(NULL, mpierr, __FILE__, __LINE__);


```

The first parameter of check_mpi() (NULL in the code above), is the file pointer, or NULL if there is none available.

The check_mpi function will print an MPI error message, then call check_netcdf with the error code PIO_EIO. This will consult the error handler set for the file (if there is a file). The error handler may call MPI_Abort() and end everything.

The PIO_EIO error code is then returned to the calling program (if MPI abort was not called by the error handler).

Here's what I am doing for netCDF error codes:

```
    /* Broadcast and check the return code. */
    if ((mpierr = MPI_Bcast(&ierr, 1, MPI_INT, ios->ioroot, ios->my_comm)))
        return check_mpi(file, mpierr, __FILE__, __LINE__);
    if (ierr)
        return check_netcdf(file, ierr, __FILE__, __LINE__);

```
In this case ierr is the return code. I broadcast it to all tasks (from ioroot), and the all tasks can call check_netcdf() if ierr is not zero.

edhartnett:
For the errors that are neither MPI nor netCDF errors, we do error handling inconsistently. Sometimes we return an error code:

  ```
  /* Find the info about this file. */
    if ((ierr = pio_get_file(ncid, &file)))
        return ierr;
```

[...]

PR #228, opened 2016-11-25
Checking return values for MPI calls in pio_spmd.c (also continued development of test_spmd)

edhartnett:
In the interests of keeping code changes separate from test changes, I would like to merge my current branch, and continue development of test_spmd on a new branch.

In this PR I added checking of return values for all MPI function calls in pio_spmd.c. (See #225). 

There is also some development of test_spmd. I am getting some confusing results, so I wanted to make sure that none of the MPI functions was returning errors.

### DIFF
--- a/tests/unit/test_spmd.c
+++ b/tests/unit/test_spmd.c
@@ -12,7 +12,7 @@
 #include <sys/time.h>
 
 /* The number of tasks this test should run on. */
-#define TARGET_NTASKS 4
+#define TARGET_NTASKS 2
 
 /* The name of this test. */
 #define TEST_NAME "test_spmd"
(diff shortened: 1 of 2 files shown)


tests/unit/test_spmd.c
```diff
@@ -52,64 +38,45 @@ int run_spmd_tests(MPI_Comm test_comm)
...
     /* Initialize the arrays. */
     for (int i = 0; i < ntasks; i++)
     {
-        sendcounts[i] = i + 1;
-        recvcounts[i] = my_rank + 1;
-        rdispls[i] = i * (my_rank + 1) * sizeof(int);
-        sdispls[i] = (((i+1) * (i))/2) * sizeof(int);
-        sendtypes[i] = recvtypes[i] = MPI_INT;
+        sendcounts[i] = 1;
+        sdispls[i] = 0;
```
jedwards4b: Displacements are in units of bytes.

PR #231, opened 2016-11-25
Finish reuse of old test code, retained unused function

### LINKED ISSUE #227: Finish rehabilitating test code in test_spmd.c
There was some test code in main() in pio_spmd. I have moved it into its own C test, test_spmd.c. Unfortunately the test code didn't work right out of the box. It needs to be fixed so that the spmd code is tested.

edhartnett:
In this PR the test_spmd.c program now uses the historical test from the main() that used to be in pio_spmd.c. This test is for the function pio_swapm(). It runs swapm for a simple case with a variety of options and checks the result.

Just as with pio_gather(), the function pio_fc_gatherv() was also not really used, because in all cases it was passes with a final parameter of 0, which causes an MPI_gatherv() call to be used instead. Now the code just calls MPI_Gatherv() directly, and pio_fc_gatherv() is gone. This is similar to the recently removed pio_fc_gather() function.

Also changed test_spmd.c so that it can handle any number of tasks between 1-4, so that it does not break the MPISERIAL run.

Fixes #227.


jedwards4b:
I don't want to do this.  The pio_fc_gatherv is a useful function that can significantly improve communications performance depending on the network hardware, removing it is a step in the wrong direction.    I would much see the parameters exposed to the user to allow tuning and to 
provide a tool that would allow the user to determine the best settings for these tuning parameters.

edhartnett:
I am in favor of leaving it in if it can provide useful optimizations.

Seems like the only time we use a gatherv is during initialization.

So where could this cause better communications performance? Only at initialization? Or later, during data reads and writes?

jedwards4b:
I think there is potential to use this in the subset rearranger during read/write - I cannot recall right now why it isnt used there.

edhartnett:
OK, let's do this.

I will restore the function.

The way it was being used in the code was just as a complicated way of calling MPI_Gatherv, so we will leave my changes there (i.e. changes to call plain old MPI_Gatherv).

I will note in the comments that there is hope to use this function for future optimizations.

Then, as a future pull request, someone can add unit test for this function, and also start using it in the code. 

Sound good?

PR #233, opened 2016-11-26
Check return values from most MPI function calls

edhartnett:
There are still some stragglers but this PR adds return code checking for most of our MPI function calls.

This is in support of #225.

### DIFF
--- a/src/clib/pioc.c
+++ b/src/clib/pioc.c
@@ -518,7 +518,8 @@ int PIOc_Init_Intracomm(const MPI_Comm comp_comm, const int num_iotasks,
     }
 
     /* Create an array that holds the ranks of the tasks to be used for IO. */
-    iosys->ioranks = calloc(sizeof(int), iosys->num_iotasks);
+    if (!(iosys->ioranks = calloc(sizeof(int), iosys->num_iotasks)))
+        return PIO_ENOMEM;
     for (int i = 0; i < iosys->num_iotasks; i++)
     {
         iosys->ioranks[i] = (base + i * ustride) % iosys->num_comptasks;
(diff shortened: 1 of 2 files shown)


src/clib/pioc.c
```diff
@@ -468,104 +468,110 @@ int PIOc_Init_Intracomm(const MPI_Comm comp_comm, const int num_iotasks,
...
+        iosys->ioranks[i] = (base + i * ustride) % iosys->num_comptasks;
+        if (iosys->ioranks[i] == iosys->comp_rank)
+            iosys->ioproc = true;
+    }
+    iosys->ioroot = iosys->ioranks[0];
+
+    for (int i = 0; i < iosys->num_iotasks; i++)
+        LOG((3, "iosys->ioranks[%d] = %d", i, iosys->ioranks[i]));
+
+    /* Create an MPI info object. Actually we don't want to do
```
jedwards4b: I think that this is incorrect - where do you make the info_create call then?

src/clib/pioc.c
```diff
@@ -468,104 +468,110 @@ int PIOc_Init_Intracomm(const MPI_Comm comp_comm, const int num_iotasks,
...
+    if ((mpierr = MPI_Group_incl(iosys->compgroup, iosys->num_iotasks, iosys->ioranks,
+                                 &iosys->iogroup)))
+        return check_mpi(NULL, mpierr, __FILE__, __LINE__);        
+
+    /* Create an MPI communicator for the IO tasks. */
+    if ((mpierr = MPI_Comm_create(iosys->comp_comm, iosys->iogroup, &iosys->io_comm)))
+        return check_mpi(NULL, mpierr, __FILE__, __LINE__);                
+
+    /* For the tasks that are doing IO, get their rank within the IO communicator. */
+    if (iosys->ioproc)
```
jedwards4b: I don't see any reason you couldn't change this one.

edhartnett: I know, it's very odd. But when I change that one, all tests break.

So I thought I would take the low-hanging fruit and get the rest of these changes merged. Then I will swing back around and investigate this one.

PR #235, opened 2016-11-28
Change test_nc4 to enable async testing of these functions...

edhartnett:
Until now, I've had separate test programs for async tests. This made sense in the old days when asynch could be turned off and on with a build option.

Now that async is always there, I would like tests to run the same test code on both async and non-async.

In this PR I have modified test_nc4.c to accomplish this. The async function still does not actually test the nc4 code - that will come in the next PR. I know that there will be changes in pio_nc4.c once I start testing, and I want to keep those changes separate.

Only logging and whitespace changes in library code in this PR. Code changes are in test_nc4.c.


rljacob:
Where is the github issue explaining the need for these changes?

edhartnett:
@rljacob I have added the issue number to the description, it is #106. 

Due to a mistake on my part, the additional functions related to netCDF-4 optimization were not tested for async. In fact, looking at the code, they certainly will not work for async. :-) I just forgot to convert them.

So I need to test them and get them working. 

All my tests until now have been either async or not async, and that is rather wasteful. Because mostly the test code is the same, only the initialization changes. So my idea is to modify the C tests so that they can test the same functionality under both async and non-async.

The changes in this PR were my first attempt, on test_nc4.c. So now this test runs both async and non-async versions of the test code, which is all now in one function so it can be easily used by both.

In this PR, the actual testing of netCDF-4 optimizations under async was turned off (non-async was left on). 

Now that this is merged, I will start another branch, with the nc4 code turned on, and will make async modifications in a separate pull request. This will involve a bunch of changes to pio_msg.c and pio_nc4.c as I put in the async feature for these functions.

In further PRs, I will apply these sorts of changes to the other C tests, so that their PIO test code is isolated in a function, and can be called by both async and non-async cases. In that way I will increase test coverage without repeating a lot of test code.

I also hope (in further PRs) to whittle down some of the existing _async tests that I have added, converting them to test both async and non-async, or else eliminating some of them entirely. I had a lot of trouble getting the async code to work, and there is currently a lot of redundant testing.

PR #236, opened 2016-11-28
Some mem cleanup

jayeshkrishna:
This commit includes several memory related cleanups in the code and the testing infrastructure. The commit also improves the pio logging infrastructure by adding an explicit init and finalize.


src/clib/pio_file.c
```diff
@@ -109,6 +109,7 @@ int PIOc_createfile(const int iosysid, int *ncidp, int *iotype,
         return PIO_ENOMEM;
 
     /* Fill in some file values. */
+    file->fh = -1;
```
edhartnett: Why are you initializing fh?

src/clib/pioc_support.c
```diff
@@ -93,19 +94,57 @@ int PIOc_strerror(int pioerr, char *errmsg)
...
+}
 
-    /* Create a filename with the rank in it. */
-    MPI_Comm_rank(MPI_COMM_WORLD, &my_rank);
-    sprintf(log_filename, "pio_log_%d.txt", my_rank);
+/** Initialize logging
+ * Open log file, if not opened yet, or
+ * increment ref count if already open
+ */
+void pio_init_logging(void )
```
edhartnett: What is the reason for doing this in Init_Intercomm or PIOc_Init_Async? Why not just leave the initialization for PIOc_set_log_level()?

The problem is that there are functions that the user might call before init_intracomm or Init_Async. What happens to log output there?

What is the reason for moving the logging code? Is this related to a github issue?

src/clib/pioc_support.c
```diff
@@ -93,19 +94,57 @@ int PIOc_strerror(int pioerr, char *errmsg)
...
+        sprintf(log_filename, "pio_log_%d.txt", my_rank);
+
+        /* Open a file for this rank to log messages. */
+        LOG_FILE = fopen(log_filename, "w");
+
+        pio_log_ref_cnt = 1;
+    }
+    else
+    {
+        pio_log_ref_cnt++;
```
edhartnett: Why do we need a ref count?

src/clib/pioc_support.c
```diff
@@ -148,35 +188,43 @@ void pio_log(int severity, const char *fmt, ...)
        many tabs before the message. */
     if (!severity)
     {
-        strcpy(ptr, ERROR_PREFIX);
+        strncpy(ptr, ERROR_PREFIX, (rem_len > 0) ? rem_len : 0);
```
edhartnett: What is this change for?

src/clib/pioc_support.c
```diff
@@ -148,35 +188,43 @@ void pio_log(int severity, const char *fmt, ...)
...
-    strcpy(ptr, rank_str);
+    snprintf(rank_str, MAX_RANK_STR, "%d ", my_rank);
+    strncpy(ptr, rank_str, (rem_len > 0) ? rem_len : 0);
     ptr += strlen(rank_str);
+    rem_len -= strlen(rank_str);
 
     /* Print out the variable list of args with vprintf. */
     va_start(argp, fmt);
-    vsprintf(ptr, fmt, argp);
+    vsnprintf(ptr, ((rem_len > 0) ? rem_len : 0), fmt, argp);
```
edhartnett: What is this change for?

src/clib/pio_file.c
jayeshkrishna: Initializing is always good IMO. Esp, for procs where it is not set.

src/clib/pioc_support.c
```diff
@@ -170,6 +170,7 @@ void pio_log(int severity, const char *fmt, ...)
 {
     va_list argp;
     int t;
+    int rem_len = MAX_LOG_MSG;
     char msg[MAX_LOG_MSG];
```
edhartnett: You want msg to be MAX_LOG_MSG + 1.

The plus one is for the null that is on the end.

src/clib/pio_file.c
edhartnett: In C we don't initialize unless there is a need.

What you are doing is communicating that you are going to depend on that value being -1. But we don't. So don't communicate that we are going to care about that value, since we don't.

I will agree that it should be removed from the log statement on line 191. There's no way it can be set by then, so it's pointless to log it.

src/clib/pioc_support.c
jayeshkrishna: The old implementation looked hacky to me. PIOc_set_log_level() should not be initializing the logging, it should be setting the log level. The logging should be initialized when one initializes the pio library.
When are the log files closed in the prior impl? It also used invalid file handles for ops.
Also, Logging is always routed to stdout (even if the log files are not open yet).
If you want to improve on this impl, please feel free to create a new PR after this change is merged.

src/clib/pioc_support.c
```diff
@@ -148,35 +188,43 @@ void pio_log(int severity, const char *fmt, ...)
...
+    {
+        strncpy(ptr++, "\t", (rem_len > 0) ? rem_len : 0);
+        rem_len--;
+    }
 
     /* Show the rank. */
-    sprintf(rank_str, "%d ", my_rank);
-    strcpy(ptr, rank_str);
+    snprintf(rank_str, MAX_RANK_STR, "%d ", my_rank);
+    strncpy(ptr, rank_str, (rem_len > 0) ? rem_len : 0);
```
edhartnett: Same comment as above. There is no way that rem_len can be 0 or less here.

src/clib/pioc_support.c
jayeshkrishna: The application can have multiple calls to pio_init() . This is the right way, IMO, to clean up internal resources in the lib (Should be adopted for freeing the buffer pool too).

src/clib/pioc_support.c
jayeshkrishna: We should not be using strcpy

src/clib/pioc_support.c
jayeshkrishna: We should not be using vsprintf

src/clib/pioc_support.c
```diff
@@ -148,35 +188,43 @@ void pio_log(int severity, const char *fmt, ...)
...
     /* Send message to log file. */
-    fprintf(LOG_FILE, "%s", msg);
+    if (LOG_FILE)
+        fprintf(LOG_FILE, "%s", msg);
 
     /* Ensure an immediate flush of stdout. */
     fflush(stdout);
-    fflush(LOG_FILE);
+    if (LOG_FILE)
+        fflush(LOG_FILE);
```
edhartnett: Is there some case in which LOG_FILE is not set when this code runs? If not, then you can check with an assert(LOG_FILE) instead of using if statements.

The if statements communicate that there is some valid execution path where LOG_FILE is NULL. Is there such a path?

src/clib/pio_file.c
jayeshkrishna: Note that logging statements in the code already assume (or don't make sense otherwise) that this handle is initialized.

src/clib/pio_darray_int.c
```diff
@@ -1527,6 +1527,7 @@ void free_cn_buffer_pool(iosystem_desc_t ios)
         cn_buffer_report(ios, true);
         bpoolrelease(CN_bpool);
         LOG((2, "free_cn_buffer_pool done!"));
+        free(CN_bpool);
```
edhartnett: Good leak catch.

src/clib/pioc_support.c
jayeshkrishna: Also, I think we should have uniform coding style through out the function (easier to modify the code later). I find it more easy to understand and modify. strcpy is waiting for a bug.

src/clib/pioc_support.c
jayeshkrishna: I think the code is self explanatory. Adding comments would be too verbose.

src/clib/pioc_support.c
edhartnett: In the future, if you wish to have less changes in review, you should file a github issue before making changes, so that we agree as to what is supposed to be happening.

I agree there are some bugs with the log files, but you have introduced a lot of unnecessary code to deal with them.

src/clib/pioc_support.c
edhartnett: The application can have multiple calls to pio_init, but there is no need to initialize the logging there. Leave the initialization in PIOc_set_log_level() and we don't need ref counts.

src/clib/pioc_support.c
edhartnett: The coding style is that we don't add unnecessary code.

You may find your own unnecessary code to be easy to understand and modify, but I do not. So let's stick to only necessary code, then we will both (and all other C programmers) have the minimum of effort to understand the code.

src/clib/pio_file.c
jayeshkrishna: In this case fh is set on some procs and not on others. Initializing it to -1 ensures that when we debug the code we know that the handle is not set.

src/clib/pioc_support.c
jedwards4b: But if multiple components call pioc_set_log_level  on the same tasks what happens?

jayeshkrishna: As I mentioned above, some issues with the older code,
* When are the log files closed (The log files were not closed AFAICT)? Log flies are internal resources used by the library and needs to be cleaned/closed up.
* There were no checks for the validity of the file handle before using it

src/clib/pioc_support.c
jayeshkrishna: There are a bunch of copies to the same buffer. Having a consistent style of checking the remaining length is easier to read.

src/clib/pioc_support.c
jayeshkrishna: There can be a future enhancement to the code where developers can choose whether to write logs to a file (just print it out vs just write to a log file vs do both)

src/clib/pioc_support.c
jayeshkrishna: Also what happens when you have a log statement that exceeds the MAX_LOG_MSG ? The right approach in this case would be to truncate the message (rather than a buffer overflow).

src/clib/pioc_support.c
rljacob: If the function is called "set log level" is should only set the log level.

src/clib/pioc_support.c
edhartnett: I don't object to a change to using strncpy. I agree that the message should be tuncated.

We cannot, obviously, make the code proof to future programming bugs. Keeping code to a minimum is a known way of reducing bugs.

src/clib/pioc_support.c
edhartnett: Please do not add code which may be needed in some future enhancement.

When that enhancement is made, it is up to that programmer to add these changes. Until then it is speculative.

src/clib/pio_file.c
jedwards4b: Yes it does help in debugging to set this to a known value on tasks where it is not used.

src/clib/pioc_support.c
edhartnett: If multiple components (or the same component multiple times) makes calls to PIOc_set_log_level, then I think it should change logging to the specified level for that task, and append to the existing log file. (And if you set it to different levels on the same task in different components, you get the last one set.)

That's not what will happen right now. At the moment, the file will be overwritten every time the log level changes, and that is clearly a bug. So let's fix that.

@rjacob the fist message logged (and this is very helpful) is the one saying that the log level has been set. This is useful because it means that the logging is working. There are also some functions that use logging that are called before pio_init. (The nc4 functions that can control netCDF-4 optimizations. Yes, I know they are a pain. But logging is still very useful there.)

So there is good reason to start the log files in the first call any task makes to PIOc_set_log_level().

The log files should be closed (if open) in PIOc_finalize(). It is a bug that they are not currently closed.

edhartnett: An un-raised issue is what happens to log files if this is run on many, many processors?

For example, if I accidentally leave logging on, and run the code on 100K tasks, I will get 100K files. That can't be good. Obviously I'm not going to look at 100K files.

Maybe we should build in some limit, like 32?

rljacob:
This is a good first-step cleanup of some code that really needs a clean up.   I recommend merging this and any problems can be fixed in future PRs.

src/clib/pioc_support.c
jedwards4b: I don't think that this point is worth arguing over.

src/clib/pioc_support.c
jayeshkrishna: The other side of this problem (log files don't capture logs before pio_init() ) is that logs before a user sets pioc_set_log_level() is not captured by the current impl. With the proposed change , as before, all logs (including the logs before user sets pioc_set_log_level()) are directed to stdout.
I think it is cleaner to have init/finalize for the logging (and that these routines manage all the logic required to build and tear up logging support). The reference count is required in this implementation to keep track of inits/finalizes.
Setting log levels should be separate from setting up the logging infrastructure (log files etc).

jayeshkrishna: A simple solution (which is what I had in mind when I was modifying this code) for a large number of log files being created for runs with a lot of procs is to allow the user to choose between (1) log on stdout and log files (2) log only on stdout -- can be used for logging in large runs (3) log only to log files

rljacob:
Re: a prefacing github issue to link to.  This is part of the general work on fixing memory leaks and hardening the code that we all agreed was a priority.   I don't think those PRs necessarily require a corresponding github issue.  Github issues should be used mostly to report bugs or request new features.

PR #253, opened 2016-11-30
cleaned up documentation

edhartnett:
Cleaned up documentation and whitespace issues. No code changes.

### DIFF
--- a/src/clib/pio_file.c
+++ b/src/clib/pio_file.c
@@ -6,12 +6,14 @@
 #include <pio.h>
 #include <pio_internal.h>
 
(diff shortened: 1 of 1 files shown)


src/clib/pio_file.c
```diff
@@ -429,7 +429,7 @@ int PIOc_deletefile(const int iosysid, const char filename[])
         MPI_Barrier(ios->io_comm);
     }
 
-    //   Special case - always broadcast the return from the
+    /* Special case - always broadcast the return from the ??? */
```
jedwards4b: Always broadcast the return from the delete command, I'm not entirely sure why at this point.

PR #259, opened 2016-12-01
Return checking for MPI functions in pio_darray_int.c

### LINKED ISSUE #257: Check MPI return codes for pio_darray_int.c
There are lots of MPI calls in pio_darray_int.c that ignore return values. They need to be checked in all cases.

edhartnett:
Fixes #257. Part of #225.

Added return code checking to MPI calls that didn't have it in pio_darray_int.c.

### DIFF
--- a/src/clib/pio_darray_int.c
+++ b/src/clib/pio_darray_int.c
@@ -62,7 +62,8 @@ void compute_buffer_init(iosystem_desc_t ios)
     LOG((2, "compute_buffer_init CN_bpool = %d", CN_bpool));
 }
 
-/** Write a single distributed field to output. This routine is only
+/**
+ * Write a single distributed field to output. This routine is only
  * used if aggregation is off.
  *
  * @param file a pointer to the open file descriptor for the file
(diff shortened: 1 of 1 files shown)


src/clib/pio_darray_int.c
```diff
@@ -374,8 +391,9 @@ int pio_write_darray_nc(file_desc_t *file, io_desc_t *iodesc, const int vid,
                         while(vdesc->request[reqn] != NC_REQ_NULL)
                             reqn++;
 
-                    ierr = ncmpi_bput_varn(file->fh, vid, rrcnt, startlist, countlist,
-                                           IOBUF, iodesc->llen, iodesc->basetype, vdesc->request+reqn);
+                    if ((ierr = ncmpi_bput_varn(file->fh, vid, rrcnt, startlist, countlist, IOBUF, iodesc->llen, iodesc->basetype, vdesc->request+reqn)))
```
jedwards4b: seems like something is missing here?  The if statement is imcomplete.

edhartnett: Thanks for catching that! Fixed...

PR #265, opened 2016-12-02
code review process changes based on github issue

### LINKED ISSUE #261: Review procedure for code review
Code review is a well-proven tool for detecting defects and improving code.

To aid the discussion here's what we have now:

<li>All code is subject to review.

<li>Pull requests will be as small as possible and focused on
one(-ish) issue.

<li>All non-trivial pull requests are associated with a github
issue. The issue is where discussion of requirements and
implementation details can be worked out.

<li>Pull requests will be left up on github for about a day. Request
more time if you need it and are actively reviewing the code. (Note
that pull request can also be reviewed after they are merged, if you
miss one).

<li>Jim will identify key stakeholders in chaged code and ensure they
accept code changes.

<li>Reviewers are open-minded and ready to accept improvements to the
library.

<li>Reviewers will make comments on the pull request. All comments
must be resolved.

<li>If chages are dictated, they happen on the branch, so code
reviewers can see the updated code.

<li>The pull request is only merged when all programmers agree that
all issues have been resolved.

edhartnett:
Fixes #261.

### DIFF
--- a/doc/source/base.txt
+++ b/doc/source/base.txt
@@ -47,5 +47,5 @@ Please review the ChangeLog that is included with the distribution for up-to-dat
  - \ref examp
  - \ref faq
  - \ref api
- - \ref code_style
+ - \ref contributing_code
 */
--- a/doc/source/contributing_code.txt
+++ b/doc/source/contributing_code.txt
@@ -94,8 +94,9 @@ correctly. M-x indent-region will indent a selected region of code.
 
 <li>All code is subject to review.
 
-<li>Pull requests will be as small as possible and focused on
-one(-ish) issue.
+<li>Pull requests will be focused on one issue.
+
+<li>Pull requests may not be submitted until all tests pass.
 
 <li>All non-trivial pull requests are associated with a github
 issue. The issue is where discussion of requirements and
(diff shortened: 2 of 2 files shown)


jayeshkrishna:
A topic of discussion is whether we want to add tests to the test suite (in a PR) even if they don't pass (the new tests don't pass).
Adding new tests (even if they don't pass) would help us in knowing the current health of the code. Future issues and PRs can fix these tests.

jedwards4b:
I don't want tests in the test suite that regularly fail, it's just too easy to ignore the status for all tests when one of them is failing.   I would prefer if the solution were added at the same time as the new test.  If that isn't possible then the test should be on a branch with an associated issue until such time as the issue is resolved, and added to the test suite at that time.   Having said that I also don't want tests removed from the test suite because a recent update broke them.

edhartnett:
The usual practice is not to have broken tests checked into master.

If I can get my hands on a jenkins server, it is easy to have it checkout PRs and run all tests. That way we know that all tests pass while reviewing the code.

rljacob:
Can the current test suite be run "by hand" from the command line?  That or something similar should be required before issuing the PR.   The Jenkins PR testing would be ideal.

Known broken features/bugs can either sit in a github issue that ages or in a test report that ages or both.  But they need to be tracked somehow as soon as they are identified.

Katetc:
The current test suite is easy to run by hand. Build the project and then run "ctest" from the base directory. Instructions here: http://ncar.github.io/ParallelIO/test.html

One stumbling point is that not everybody has access to the same machines. Several collaborators do not have accounts on yellowstone, for example. So, a Jenkins auto-tester would be helpful.

gold2718:
I agree with @jedwards4b that the new test should come in with the PR that fixes the bug.

Issue #238, opened 2016-11-29
Setting error handler not currently working with async


edhartnett:
Currently setting the error handler does not work with async.

PR #272, opened 2016-12-04
adding pio_err() function to handle PIO errors

edhartnett:
Working on #238, need to clear #225.

In this PR I add pio_err(), the function that will handled errors in the PIO library. It checks with the file and iosystem error handlers (in that order) to decide what to do with the error.

Also I note that setting the file error_handler actually just changes the iosystem error handler for that file, which is not right. So I have added an error_handler field to file_desc_t, it is initialized to 0 when a new file_desc_t is created.

### DIFF
--- a/src/clib/pio_internal.h
+++ b/src/clib/pio_internal.h
@@ -77,7 +77,7 @@ extern "C" {
         bool isend;
     } pio_swapm_defaults;
 
-    /* Handle are error in the PIO library. */
+    /* Handle an error in the PIO library. */
     int pio_err(iosystem_desc_t *ios, file_desc_t *file, int err_num, const char *fname,
                 int line);
     
--- a/src/clib/pio_internal.h
+++ b/src/clib/pio_internal.h
@@ -77,6 +77,10 @@ extern "C" {
         bool isend;
     } pio_swapm_defaults;
 
+    /* Handle are error in the PIO library. */
+    int pio_err(iosystem_desc_t *ios, file_desc_t *file, int err_num, const char *fname,
+                int line);
+    
     void pio_get_env(void);
     int  pio_add_to_iodesc_list(io_desc_t *iodesc);
     io_desc_t *pio_get_iodesc_from_id(int ioid);
--- a/src/clib/pioc_support.c
+++ b/src/clib/pioc_support.c
@@ -409,6 +409,79 @@ int check_netcdf(file_desc_t *file, int status, const char *fname, const int lin
 }
 
 /**
+ * Handle an error in PIO. This will consult the error handler
+ * settings and either call MPI_Abort() or return an error code.
+ *
+ * The error hanlder has three settings:
+ *
+ * Errors cause abort: PIO_INTERNAL_ERROR.
+ *
+ * Error codes are broadcast to all tasks: PIO_BCAST_ERROR.
+ *
+ * Errors are returned to caller with no internal action:
+ * PIO_RETURN_ERROR.
+ *
+ * @param ios pointer to the IO system info. Ignored if NULL.
+ * @param file pointer to the file description data. Ignored if
+ * NULL. If not NULL, then file level error handling is applied before
+ * IO system level error handling.
+ * @param err_num the error code
+ * @param fname name of code file where error occured.
+ * @param line the line of code where the error occurred.
+ * @returns err_num if abort is not called.
+ */
+int pio_err(iosystem_desc_t *ios, file_desc_t *file, int err_num, const char *fname,
+            int line)
+{
+    char err_msg[PIO_MAX_NAME + 1];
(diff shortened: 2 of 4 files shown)


src/clib/pio_internal.h
```diff
@@ -77,6 +77,10 @@ extern "C" {
         bool isend;
     } pio_swapm_defaults;
 
+    /* Handle are error in the PIO library. */
```
jedwards4b: Handle an error

src/clib/pioc_support.c
```diff
@@ -409,6 +409,79 @@ int check_netcdf(file_desc_t *file, int status, const char *fname, const int lin
...
+ */
+int pio_err(iosystem_desc_t *ios, file_desc_t *file, int err_num, const char *fname,
+            int line)
+{
+    char err_msg[PIO_MAX_NAME + 1];
+    int err_handler = PIO_INTERNAL_ERROR; /* Default error handler. */
+    int ret;
+    
+    /* User must provide this. */
+    pioassert(fname, "file name must be provided", __FILE__, __LINE__);
```
jedwards4b: Aren't there cases where an error may occur on a non file related issue?

src/clib/pioc_support.c
```diff
@@ -409,6 +409,79 @@ int check_netcdf(file_desc_t *file, int status, const char *fname, const int lin
...
+
+    /* Get the error message. */
+    if ((ret = PIOc_strerror(err_num, err_msg)))
+        return ret;
+
+    /* If logging is in use, log an error message. */
+    LOG((0, "%s err_num = %d fname = %s line = %d", err_msg, err_num, fname ? fname : '\0', line));
+
+    /* ??? */
+    print_trace(stderr);
```
jedwards4b: This is for debugging only and will print a traceback of the call tree.  Should only be called in the case of pio_internal_error

src/clib/pioc_support.c
```diff
@@ -409,6 +409,79 @@ int check_netcdf(file_desc_t *file, int status, const char *fname, const int lin
...
+        err_handler = file->error_handler ? file->error_handler : file->iosystem->error_handler;
+    else if (ios)
+        err_handler = ios->error_handler;
+
+    /* Should we abort? */
+    if (err_handler == PIO_INTERNAL_ERROR)
+    {
+#ifdef MPI_SERIAL
+        /* Why the special abort() call for MPI_SERIAL??? */
+        abort();
```
jedwards4b: Not sure why - I just checked the mpi-serial library and there is an abort function.

src/clib/pioc_support.c
```diff
@@ -409,6 +409,79 @@ int check_netcdf(file_desc_t *file, int status, const char *fname, const int lin
...
+#else
+        MPI_Abort(MPI_COMM_WORLD, -1);
+#endif
+    }
+
+    /* What should we do here??? */
+    if (err_handler == PIO_BCAST_ERROR)
+    {
+        /* ??? */
```
jedwards4b: Broadcast the error message to all calling tasks.  That would be all compute tasks in ios.   I can't think of a non-fatal error from pio that didn't originate in netcdf so I'm not sure of a case of a non-fatal error for which this mode would be useful.

src/clib/pioc_support.c
edhartnett: I have added an issue to investigate its removal, after all the error handling stuff is working and well-tested.

src/clib/pioc_support.c
edhartnett: OK I've documented it and moved it to the pio_internal_error block.

src/clib/pioc_support.c
edhartnett: That's the file parameter. The fname parameter is the name of the code file from the __FILE__ macro. It is always present.

The file parameter may be null, for example if the error occurs before a file is created. In this case the file parameter will be ignored.

src/clib/pio_internal.h
edhartnett: Fixed.

src/clib/pioc_support.c
edhartnett: For example you do a PIOc_createfile() with an invalid iosystemid. That would cause an error, but you may just want to return to the calling program on the compute nodes, instead of calling MPI abort.

But I am not sure how to accomplish this. I can't think of any case where the error would not be returned on all compute nodes already. For example if all compute nodes call PIOc_createfile() with a bad iosystemid, then they will already all get the error code PIO_EBADID, without any additional effort on our part.

If only one compute node calls PIOc_createfile() with a bad ID, and the rest call with a good ID, how am I supposed to distribute that error?

edhartnett:
I am going to transfer the discussion about PIO_BCAST_ERROR to the github issue.

I am going to merge these changes.

Then I will create a develop branch and we will transfer development for that. Further error handler work will continue there.

PR #275, opened 2016-12-05
fixed returns in pio_file.c to use pio_err(), working out code merge procedure

edhartnett:
A very simple PR to take us through our first merge with the develop branch.

This PR uses the pio_err() function in the rest of pio_file.c. (See issue #225.)

I have also started to document our merge procedure, which is being discussed in email and other place. Please take a look at what I have put in the contributions document.

### DIFF
--- a/doc/source/contributing_code.txt
+++ b/doc/source/contributing_code.txt
@@ -145,8 +145,8 @@ test cycles. (???)
 hours).
 
 <li>After all jenkins and Cdash builds complete successfully, with all
-tests passing, and no warnings, the PR is merged into master by Jim
-(???)
+tests passing, and no warnings, the PR is merged into master by the
+integrator.
 
 <li>Multiple PRs may be merged to master between test cycles. (???)
 
--- a/doc/source/contributing_code.txt
+++ b/doc/source/contributing_code.txt
@@ -136,7 +136,16 @@ develop.
 <li>When code review is complete, and the changes are approved, the PR
 is merged into the develop branch.
 
-<li> ???
+<li>The develop branch is tested automatically by Jenkins.
+
+<li>The develop branch is tested periodically by CDash (every ~6
+hours).
+
+<li>After all jenkins and Cdash builds complete successfully, with all
+tests passing, and no warnings, the PR is merged into master by Jim
+(???)
+
+<li>The branch is then deleted by whomever merged it to master.
 
 </ul>
 
--- a/doc/source/contributing_code.txt
+++ b/doc/source/contributing_code.txt
@@ -131,10 +131,13 @@ all issues have been resolved.
 <li>Programmers begin work on a feature or fix by branching from
 develop.
 
-<li>When branch is ready, it is submitted to code review.
+<li>When a branch is ready, it is submitted to code review.
 
 <li>When code review is complete, and the changes are approved, the PR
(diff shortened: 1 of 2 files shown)


doc/source/contributing_code.txt
```diff
@@ -131,11 +131,14 @@ all issues have been resolved.
 <li>Programmers begin work on a feature or fix by branching from
 develop.
 
-<li>When branch is ready, it is submitted to code review.
+<li>When a branch is ready, it is submitted to code review.
```
rljacob: A single "integrator" is assigned to handle merging the PR.  Any developer my comment on a PR but only the integrator may decide to merge or not.  Developers should not integrate their own PRs.

doc/source/contributing_code.txt
```diff
@@ -131,11 +131,14 @@ all issues have been resolved.
...
 
 <li>When code review is complete, and the changes are approved, the PR
 is merged into the develop branch.
 
+<li>Mutliple merges into the develop branch may take place between
+test cycles. (???)
+
```
rljacob: Ideally it would be one at a time because that's the best way to isolate the cause of test fails.  But groups of small PRs can go in at the same time.  Major PRs that touch lots of files and change behavior should go in by themselves.  Its a judgement call.

doc/source/contributing_code.txt
```diff
@@ -131,11 +131,14 @@ all issues have been resolved.
...
 <li>The develop branch is tested automatically by Jenkins.
 
 <li>The develop branch is tested periodically by CDash (every ~6
```
rljacob: Does Cdash run tests?  I thought our setup was that Jenkins (at NCAR) runs the tests and sends results to cdash.

doc/source/contributing_code.txt
```diff
@@ -145,8 +148,17 @@ hours).
 tests passing, and no warnings, the PR is merged into master by Jim
 (???)
```
rljacob: "merged to master by the integrator".  Doesn't have to be Jim unless he wants that role.

doc/source/contributing_code.txt
```diff
@@ -145,8 +148,17 @@ hours).
...
 
+<li>Commit comments are squished, to come comment per merged PR.
```
rljacob: "to come comment"?   When a branch is part of a PR, the commits in it should already be cleaned up and ready for public consumption.   The merge to develop/master should be non-fast-forward.

doc/source/contributing_code.txt
edhartnett: Since we are going to wait for the cdash tests (which are the tests on yellowstone and other actual HPC platforms), then we can only integrate to master once per cdash testing cycle. (Every 6 hours, I believe.)

Since I routinely submit 2-4 PRs a day, the math does not work. Either we integrate multiple PRs routinely, or else I have fewer PRs (and therefor less isolation of changes in different PRs).

So I think the best way is to do multiple PRs at a time. If something fails I will just have to roll back all of them and figure it out.

doc/source/contributing_code.txt
jedwards4b: Kate and I run ctest on several platforms in cron jobs, these checkout the default branch and so the next test cycle should reflect the develop branch.

PR #277, opened 2016-12-05
fixed problem with uninitialized file_desc_t pointer

### LINKED ISSUE #276: pio_unit_test failure
It is observed that pio_unit_test fails recently, which  is related to an assertion failure:
pioc_support.c:366: check_netcdf: Assertion `file' failed.

In PIOc_deletefile(), pio_file.c, there is a pointer of file_desc_t type, which is define but never initialized or assigned:
file_desc_t *file;

Therefore, with different compilers or environments, this pointer can be a junk value or NULL.

The failure seems to be introduced by commit 1372ba5, which added a check like below:
check_netcdf(file, ierr, \__FILE\__, \__LINE\__);
When the file pointer happens to be NULL, the assertion in check_netcdf() fails.

Not all builders can catch this failure caused by uninitialized/unassigned variable: 
On CDash, it only fails for gcc 4.8.5 and PGI 16.5
On ANL Jenkins server, it only fails for valgrind check builder (not for coverage builder)

Suggestion:
1) This pointer should be explicitly initialized to NULL, or assigned with a valid value.
2) check_netcdf() should not be called if this pointer is NULL.

edhartnett:
Fixes #276.

### DIFF
--- a/src/clib/pio_file.c
+++ b/src/clib/pio_file.c
@@ -389,7 +389,6 @@ int PIOc_closefile(int ncid)
 int PIOc_deletefile(int iosysid, const char *filename)
 {
     iosystem_desc_t *ios;  /* Pointer to io system information. */
-    file_desc_t *file;     /* Pointer to file information. */
     int ierr = PIO_NOERR;  /* Return code from function calls. */
     int mpierr = MPI_SUCCESS, mpierr2;  /* Return code from MPI function codes. */
     int delete_called = 0; /* Becomes non-zero after delete is called. */
@@ -421,9 +420,9 @@ int PIOc_deletefile(int iosysid, const char *filename)
 
         /* Handle MPI errors. */
         if ((mpierr2 = MPI_Bcast(&mpierr, 1, MPI_INT, ios->comproot, ios->my_comm)))
-            return check_mpi(file, mpierr2, __FILE__, __LINE__);
+            return check_mpi(NULL, mpierr2, __FILE__, __LINE__);
         if (mpierr)
-            return check_mpi(file, mpierr, __FILE__, __LINE__);
+            return check_mpi(NULL, mpierr, __FILE__, __LINE__);
     }
 
     /* If this is an IO task, then call the netCDF function. The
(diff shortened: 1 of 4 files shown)


src/clib/pio_file.c
```diff
@@ -452,9 +451,9 @@ int PIOc_deletefile(int iosysid, const char *filename)
 
     /* Broadcast and check the return code. */
     if ((mpierr = MPI_Bcast(&ierr, 1, MPI_INT, ios->ioroot, ios->my_comm)))
-        return check_mpi(file, mpierr2, __FILE__, __LINE__);        
+        return check_mpi(NULL, mpierr2, __FILE__, __LINE__);        
     if (ierr)
-        return check_netcdf(file, ierr, __FILE__, __LINE__);
+        return check_netcdf(NULL, ierr, __FILE__, __LINE__);
```
jayeshkrishna: Doesn't check_netcdf assume file (ptr to file_desc_t) is valid?

edhartnett: Correct. The problem is that check_netcdf(), like pio_err(), needs to take both an iosystem_desc_t pointer and a file_desc_t pointer, in case there is a valid iosystem, but not an open file.

I am preparting a new version of check_netcdf(), and will check in my changes later this morning.

edhartnett:
Stand by on this PR until the one I just put up passes review and is merged, then I will merge it into this branch and continue work on this PR.

src/clib/pio_file.c
dqwu: A new version of check_netcdf() is needed. I checked out develop branch and merged this PR for testing, pio_unit_test still failed due to an assertion failure.
pio_unit_test: src/clib/pioc_support.c:366: check_netcdf: Assertion `file' failed.

dqwu: The failure is "invalid memory reference" caused by dereferencing a NULL pointer. A builder configured as optimize (assertions are disabled) will also fail.

edhartnett:
OK, try again.

There is now a new function, check_netcdf2() which takes the ios.

This passes existing tests, but I still need to figure out and test what is supposed to happen to PIO_BCAST_ERROR.

I would like to merge this PR into develop, then master, to get the build working again. Then we will change our process a bit, and I will start a new branch to continue the error handling work.

PR #288, opened 2016-12-07
Untangle git situation and restart process

edhartnett:
OK, I am getting a bit tangled up in git. I merged a bunch of branches to develop, and then deleted them (out of habit). Now I realize that I need to merge them to master.

What I also realize is that I already have lost track of what PRs I have merged to develop, so merging them to master in the correct order is going to be hard.

Also, although I have fixed my problems on most platforms, the code is still failing in a few cdash tests. Doh!

So what I would like to do is a pull request from develop to master, to get master synced with develop. I know this is not in the rules, but I want to get back to where we should be, and try again.

Then I will try using git the way that Robert suggests.


rljacob:
Its possible to recover a deleted branch in git:
https://confluence.atlassian.com/bbkb/how-to-restore-a-deleted-branch-765757540.html

Issue #284, opened 2016-12-06
Fortran testing framework does not return error if a customized test driver fails


dqwu:
Simple steps to reproduce it on latest master branch:
a) Edit pio_file.c, add a line of testing code before line 190.
file->iotype = 5; /* bad iotype for testing \*/
switch (file->iotype) /* now becomes line 191 */
Note: this will intentionally make PIOc_createfile() call iotype_error() and return an error code PIO_EBADIOTYPE (-500)

b) Build tests and run "ctest -R pio_unit_test"
The test will print out some error messages and call MPI_Abort, but still be marked as "Test Passed"

testing PIO_createfile...
ERROR: iotype 5 not defined in build src/clib/pio_file.c 222
...
ERROR: iotype 5 not defined in build src/clib/pioc_support.c 410
application called MPI_Abort(MPI_COMM_WORLD, 0) - process 0
...
Test Passed.

It is confirmed that this error code is returned to the caller in basic_tests.F90
ret_val = PIO_createfile(pio_iosystem, pio_file, iotype, filename)
if (ret_val .ne. PIO_NOERR) then
...

All affected C tests will fail as expected. If we run "ctest -R test_darray":
ERROR: iotype 5 not defined in build src/clib/pio_file.c 222
...
ERROR: iotype 5 not defined in build src/clib/pioc_support.c 410
0 Error -500 in tests/unit/test_darray.c, line 126
...
Test Failed.

There are two other Fortran tests that have the same issue: ncdf_simple_tests and ncdf_fail
Even the Test driver is marked as "FAILED" in the log, the test is marked as "Passed"
...
ERROR: iotype 5 not defined in build src/clib/pioc_support.c 410
            PIO_TF:                               Test driver  ------  FAILED
\<end of output>
...
Test Passed.

Why should these Fortran tests pass despite of printed error messages and failed Test driver? Is this an expected behavior or a test framework bug?

Note:
It will be helpful if these affected Fortran tests fail as expected (that is the purpose of unit tests), otherwise it will be hard to be noticed on CDash (the tests pass anyway). Though we reproduce this issue with artificial code (set iotype to 5), in real testing iotype can have bad values. In issue #276, the uninitialized file_desc_t type pointer (file) might contain a random junk value like 0x4078a4 (which is known as a wild pointer), and the member variable file->iotype might contain a bad value like 1238417607.

jayeshkrishna:
[...]

PR #298, opened 2016-12-07
Make sure a test driver fail is signaled as a fail

jayeshkrishna:
If the test driver fails make sure that we report an error. Without
this fix failures in the test driver is not signaled as an error.

Fixes Issue #284

### DIFF
--- a/tests/general/util/pio_tf_f90gen.pl
+++ b/tests/general/util/pio_tf_f90gen.pl
@@ -623,6 +623,7 @@ sub get_default_test_main
   $out_line = $out_line . "          WRITE(*,PIO_TF_TEST_RES_FMT) \"PIO_TF: \",&\n";
   $out_line = $out_line . "           \"All tests\", \"---------\", \"PASSED\"\n";
   $out_line = $out_line . "        ELSE\n";
+  $out_line = $out_line . "          pio_tf_nerrs_total_ = pio_tf_nerrs_total_ + 1\n";
   $out_line = $out_line . "          WRITE(*,PIO_TF_TEST_RES_FMT) \"PIO_TF: \",&\n";
   $out_line = $out_line . "           \"Test driver\", \"---------\", \"FAILED\"\n";
   $out_line = $out_line . "        END IF\n";


jayeshkrishna:
Merged to develop (missed adding the PR # in commit - that would have flagged the commit here)

dqwu:
I checked out the latest develop branch and added "file->iotype = 5;" to pio_file.c (before line 190)
Now both "ctest -R ncdf_simple_tests" and "ctest -R ncdf_fail" failed as expected:
ERROR: iotype 5 not defined in src/clib/pio_file.c 222
ERROR: iotype 5 not defined in src/clib/pioc_support.c 410
...
PIO_TF: Test driver  ------  FAILED
...
Test Failed.

However, "ctest -R pio_unit_test" still passed:
ERROR: iotype 5 not defined in build src/clib/pio_file.c 222
ERROR: iotype 5 not defined in build src/clib/pioc_support.c 410
...
ret_val = -500
...
application called MPI_Abort(MPI_COMM_WORLD, 0) - process 1
...
Test Passed.

Could you please take a look? It seems that pio_unit_test does not use a test driver like ncdf_simple_tests, so the current fix in this PR has no effect on pio_unit_test?

jayeshkrishna:
pio_unit_test does not use the fortran testing framework. Also the errors that you see in the log must be the test testing fail scenarios (does the lib return an error code for these fail cases?)

jayeshkrishna:
This should be ready to be merged to master, right (cdash - nwsc-caldera07 / nwscla-laramie1- does not show fortran test failures)?

dqwu:
It is OK to be merged to master (should not break existing code). However, it would be better to look at  pio_unit_test which still passes despite of errors.

jayeshkrishna:
The pio_unit_test issue would be an issue different from this one (pio_unit_test does not use the fortran testing framework). If you see that it succeeds where it is supposed to fail can you create a new issue for that?

dqwu:
OK we can create a new issue for pio_unit_test, which does not use the Fortran test driver. The PR can be merged to master, thanks.

PR #303, opened 2016-12-08
Ejh check netcdf2

edhartnett:
Here are changes to use the new check_netcdf() function, and remove the old one.

The new function, as you may recall, accepts the iosystem_desc_t for those cases where we have an ios but not a file to decide what error handling to use.

### DIFF
--- a/src/clib/pio_getput_int.c
+++ b/src/clib/pio_getput_int.c
@@ -110,9 +110,9 @@ int PIOc_get_vars_tc(int ncid, int varid, const PIO_Offset *start, const PIO_Off
          * user may have passed in NULLs.) */
         /* Allocate memory for these arrays, now that we know ndims. */
         if (!(rstart = malloc(ndims * sizeof(PIO_Offset))))
-            return check_netcdf(ios, file, ierr, __FILE__, __LINE__);
+            return check_netcdf(ios, file, PIO_ENOMEM, __FILE__, __LINE__);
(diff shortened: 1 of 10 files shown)


src/clib/pio_getput_int.c
```diff
@@ -473,24 +473,24 @@ int PIOc_put_vars_tc(int ncid, int varid, const PIO_Offset *start, const PIO_Off
...
                 if ((ierr = PIOc_inq_dimlen(ncid, dimid[vd], &dimlen[vd])))
-                    return check_netcdf(file, ierr, __FILE__, __LINE__);
+                    return check_netcdf(ios, file, ierr, __FILE__, __LINE__);
                 LOG((3, "dimlen[%d] = %d", vd, dimlen[vd]));
             }
         }
 
         /* Allocate memory for these arrays, now that we know ndims. */
         if (!(rstart = malloc(ndims * sizeof(PIO_Offset))))
-            return check_netcdf(file, PIO_ENOMEM, __FILE__, __LINE__);
```
jedwards4b: I didn't realize that these malloc issues were being piped through a routine called check_netcdf - isn't that a bit confusing?

edhartnett: That is a good point. Let me ponder that. It would look more natural to use pio_err() for malloc errors. Let me investigate and get back to you on that.

PR #311, opened 2016-12-09
Runcdash fix mpiserial

jedwards4b:
Line to cd to src directory was missing.


jedwards4b:
I've pushed to develop.

edhartnett:
OK, sorry, looks like this was a result of my git confusion.

PR #314, opened 2016-12-12
fixed test_names problem with lack of return code for check_strerror 

### LINKED ISSUE #304: test_names failed on Jenkins
https://jenkins-ci.mcs.anl.gov/job/ParallelIO_coverage/52/console

The failure was introduced by PR #280.

The newly added function below does not have a return statement at the end
int check_strerror(int my_rank)
{
int ret;
...
}

The returned value is undefined (dependent on the compiler). For code coverage builder, it returns a non-zero value like 1, for other builders they return 0 (the test would pass).

This is similar to issue #163 we fixed before. It looks like code coverage builder is able to find this kind of errors when the return value of a function is undefined.

We need add code coverage builder on develop branch to Jenkins. This way the master branch can be protected from similar bugs in the future.

edhartnett:
Thought this fix made it in last week, but it got lost in my git confusion.

So here it is again. ;-)

Fixes #304.

### DIFF
--- a/tests/unit/test_names.c
+++ b/tests/unit/test_names.c
@@ -37,8 +37,7 @@ int dim_len[NDIM] = {NC_UNLIMITED, X_DIM_LEN, Y_DIM_LEN};
  * @param my_rank rank of process
  * @param ncid ncid of open netCDF file
  * @returns 0 for success, error code otherwise. */
-int
-check_dim_names(int my_rank, int ncid, MPI_Comm test_comm)
+int check_dim_names(int my_rank, int ncid, MPI_Comm test_comm)
 {
     char dim_name[NC_MAX_NAME + 1];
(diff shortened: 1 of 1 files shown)


tests/unit/test_names.c
```diff
@@ -274,15 +271,24 @@ int check_strerror_pio(int my_rank)
 int check_strerror(int my_rank)
 {
     int ret;
+
+    printf("checking strerror for netCDF-classic error codes...\n");
```
jedwards4b: Should all of these printf's be here?  That's going to be pretty noisy isn't it?

edhartnett: These are only in test output, so not really a problem I think.

edhartnett:
I have merged this to develop. Will merge to master tomorrow morning, if CDash tests pass tonight.

PR #321, opened 2016-12-12
cleaned up file_desc_t initialization with a calloc

### LINKED ISSUE #273: Should use calloc to get memory for file_desc_t, clean up initialization...
A helpful and common C convention is to use 0 as the default initialization value for struct members, and then use calloc for memory allocation.

Then you save the lines of code like this:

    file->buffer.validvars = 0;
    file->buffer.data = NULL;
    file->buffer.next = NULL;
    file->buffer.vid = NULL;
    file->buffer.ioid = -1;
    file->buffer.frame = NULL;
    file->buffer.fillvalue = NULL;

In the above code, only one line (setting the ioid) is needed if calloc is used instead of malloc() to get memory for the file_desc_t, because calloc will init the other values to 0, which is the same as NULL.

This can be done in PIOc_createfile() to eliminate some code and clean up a bit.

edhartnett:
Fixes #273.

Clean up if file_desc_t initialization with calloc.

### DIFF
--- a/examples/c/CMakeLists.txt
+++ b/examples/c/CMakeLists.txt
@@ -32,20 +32,17 @@ SET(SRC example1.c)
 ADD_EXECUTABLE(example1 ${SRC})
 TARGET_LINK_LIBRARIES(example1 pioc)
 
-SET(SRC example2.c)
-ADD_EXECUTABLE(example2 ${SRC})
-TARGET_LINK_LIBRARIES(example2 pioc)
-
 #===== MPE =====
 find_package (MPE "2.4.8" COMPONENTS C)
 if (MPE_C_FOUND)
+  SET(SRC example2.c)
+  ADD_EXECUTABLE(example2 ${SRC})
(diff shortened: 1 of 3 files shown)


examples/c/CMakeLists.txt
```diff
@@ -32,20 +32,17 @@ SET(SRC example1.c)
...
+  SET(SRC example2.c)
+  ADD_EXECUTABLE(example2 ${SRC})
+  TARGET_LINK_LIBRARIES(example2 pioc)
+
     target_include_directories (example2
         PUBLIC ${MPE_C_INCLUDE_DIRS})
     target_compile_definitions (example2
         PUBLIC HAVE_MPE)
     target_link_libraries (example2 pioc lmpe mpe pthread m)
-else ()
```
jedwards4b: Are you removing this test if MPE is not found?

edhartnett: Yes.

PR #323, opened 2016-12-13
free MPI communicator newcomm

wkliao:
because of the following MPI warning message.

In direct memory block for handle type COMM, 1 handles are still allocated
...


edhartnett:
This test is moving into the cunit directory, so not sure your merge will work now...

wkliao:
I assume you meant cunit in develop branch.
(This pull request is for master branch.) Luckily, the patch has only 3 lines.

PR #324, opened 2016-12-13
Moved C unit tests into their own directory.

### LINKED ISSUE #317: Split C and Fortran unit tests into different directories
The C and Fortran unit tests are mixed together, which makes the build files needlessly complicated.

By putting each in their own directory we will get two simple build files instead of one complex one, and the C-only build will be simplified.

edhartnett:
Fixes #317.

In this PR I split the C tests into their own directory. 

Now tests/unit contains fortran tests, and C tests are all in tests/cunit.


edhartnett:
I am going to merge this into develop and see how it does.

PR #327, opened 2016-12-13
added check_mpi that accepts ios, for use in deletefile function

edhartnett:
There is one case (at least) where there is an ios but not a file pointer, and that is in deletefile.

This PR adds a check_mpi2() which can be used in those cases to make sure the correct error handler is used.


edhartnett:
I will merge these into develop...

PR #335, opened 2016-12-15
silence compile warnings caused by the use of qsort

wkliao:
Got the following compile-time warning messages.

src/gptl/gptl.c: In function 'merge_thread_data':
src/gptl/gptl.c:2578: warning: passing argument 4 of 'qsort' from incompatible pointer type
/usr/include/stdlib.h:761: note: expected '__compar_fn_t' but argument is of type 'int (*)(const char **, const char **)'
src/gptl/gptl.c:2625: warning: passing argument 4 of 'qsort' from incompatible pointer type
/usr/include/stdlib.h:761: note: expected '__compar_fn_t' but argument is of type 'int (*)(const char **, const char **)'
src/gptl/gptl.c:2648: warning: passing argument 4 of 'qsort' from incompatible pointer type
/usr/include/stdlib.h:761: note: expected '__compar_fn_t' but argument is of type 'int (*)(const char **, const char **)'
src/gptl/gptl.c: In function 'collect_data':
src/gptl/gptl.c:2802: warning: passing argument 4 of 'qsort' from incompatible pointer type
/usr/include/stdlib.h:761: note: expected '__compar_fn_t' but argument is of type 'int (*)(const char **, const char **)'
/src/gptl/gptl.c:2803: warning: passing argument 4 of 'qsort' from incompatible pointer type
/usr/include/stdlib.h:761: note: expected '__compar_fn_t' but argument is of type 'int (*)(const char **, const char **)'
src/gptl/gptl.c:2845: warning: passing argument 4 of 'qsort' from incompatible pointer type
/usr/include/stdlib.h:761: note: expected '__compar_fn_t' but argument is of type 'int (*)(const char **, const char **)'

### DIFF
--- a/src/gptl/gptl.c
+++ b/src/gptl/gptl.c
@@ -195,8 +195,8 @@ static int update_ll_hash (Timer *, const int, const unsigned int);
 static inline int update_ptr (Timer *, const int);
 static int construct_tree (Timer *, Method);
 
-static int cmp (const char **, const char **);
-static int ncmp (const char **, const char **);
+static int cmp (const void *, const void *);
+static int ncmp (const void *, const void *);
 static int get_index ( const char *, const char *);
 
 typedef struct {
@@ -2933,9 +2933,9 @@ int get_index( const char * list,
 ** cmp: returns value from strcmp. for use with qsort
 */
 
-static int cmp(const char **x, const char **y)
+static int cmp(const void *x, const void *y)
 {
-  return strcmp(*x, *y);
+  return strcmp(*(char**)x, *(char**)y);
 }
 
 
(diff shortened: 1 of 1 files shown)


wkliao:
Well, I ran "make check" without encountering an error (on master branch).
Not sure if this is sufficient. I have been using qsort that way for a long time.

jedwards4b:
I would say that the untested code without warnings is an improvement over the untested code with warnings.   Thanks.

edhartnett:
Better still is cleaned up warnings *AND* a test.

@wkliao if you will put together a very simple test program, either in C or fortran, I will add a directory for timing tests, and make sure it is part of the build.

Tests never happen until some conscientious programmer takes the time to write them.

wkliao:
I am not aware that functions I changed in gptl.c are not used in PIO tests, even when PIO_ENABLE_TIMING is ON. Am I missing the tests somewhere that are for GPTL features? 

I poked around some files and discovered the followings.
1. added a call to `GPTLpr_summary()` before freeing test_comm to `pio_test_finalize()` in `tests/cunit/test_common.c`
2. tested test_darray and it ran OK (timing.summary shows zeros max and min wall times)
3. tested test_nc4 and got error message of "GPTLallocate: malloc failed for 0 bytes" from `merge_thread_data()` due to `nthreads == 0`. After changing the if condition from `nthreads == 1` to `nthreads <= 1`, the program hung in `GPTLpr_summary_file()` at the call to `collect_data()`. I guess `collect_data()` does not handle well the case when there is zero timer and with only main thread.

The codes in `collect_data()` appears kind of complicated.

edhartnett:
Well shoot! Now I'm sorry I mentioned testing. ;-)

Why don't you just check in your original change. I believe I will not be using the timing library, so I am not going to worry about testing it.

jedwards4b:
Pushing to develop

edhartnett:
I am going to merge these changes to master, and then recreate the develop branch.

PR #339, opened 2016-12-15
fix setting of iosystem error handler with no file_desc_t present for non-async cases

### LINKED ISSUE #268: Setting IOsystem error handler does not work for non-async open of missing file
As demonstrated by test_nc4, the setting of error handling for the io system is not working correctly. 

Not sure whether the problem is specific to this case or more general.

### LINKED ISSUE #330: Clean up double implementation of check_netcdf()
There are currently two implementations of check_netcdf, but we only need one implementation.

The check_netcdf() function should be change to call the new check_netcdf2() function, which optionally accepts ios pointer.

This change was made in some code that got messed up in the merging, so I am redoing it.

edhartnett:
Fixes #330.

Fixes #268.

Part of #238, part of #225.

Setting the iosystem error handler was not working (even for non-async cases). This was tested in test_nc4, where I delete a file, and then try to reopen it, only to get an error code.

This was causing an abort, because it was not accepting my setting of the iosystem error handler to RETURN. (Which I need for this test to work.)

I have introduced new functions PIOc_set_file_error() and PIOc_set_iosystem_error() because I need all API functions to return an error code on failure, and the existing PIOc_Set_File_Error_Handling() and PIOc_Set_IOSystem_Error_Handling() return the previous setting and have no way to return errors.

The old functions are still there and still supported, but I guess they should be removed, and the new ones used from the fortran API.

### DIFF
--- a/src/clib/pioc.c
+++ b/src/clib/pioc.c
@@ -55,12 +55,16 @@ int PIOc_File_is_Open(int ncid)
 }
 
 /**
- * Set the error handling method to be used for subsequent pio
- * library calls, returns the previous method setting.
+ * Set the error handling method to be used for subsequent pio library
+ * calls, returns the previous method setting. This function is
+ * supported but deprecated. New code should use
+ * PIOc_set_file_error().  This method has no way to return an error,
+ * so any failure will result in MPI_Abort.
  *
  * @param ncid the ncid of an open file
  * @param method the error handling method
- * @returns 0 on success, error code otherwise
+ * @returns old error handler
+ * @ingroup PIO_error_method
  */
 int PIOc_Set_File_Error_Handling(int ncid, int method)
 {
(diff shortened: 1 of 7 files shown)


src/clib/pioc.c
```diff
@@ -55,26 +55,66 @@ int PIOc_File_is_Open(int ncid)
...
+/**
+ * Set the error handling method to be used for subsequent pio
+ * library calls on this file.
+ *
+ * @param ncid the ncid of an open file
+ * @param method the error handling method
+ * @returns 0 for success, error code otherwise.
+ * @ingroup PIO_error_method
+ */
+int PIOc_set_file_error(int ncid, int method)
```
jedwards4b: I think I don't like this name - it implies you are setting an error when you are setting an error handling method.

jedwards4b: It's also useful and used in many users codes to know what the previous method was so that it can be
reset when the user is done using the new method - if it's not in the return code it should be in the arguments.

edhartnett:
Made changes in response to review feedback from @jedwards4b

edhartnett:
I have merged this with develop for testing.

PR #342, opened 2016-12-16
Fixing memory leaks on several C unit tests

dqwu:
A few C unit tests (test_darray, test_file, test_names, test_nc4) have memory leaks due to missing necessary PIOc_finalize() calls. This fix will be merged to the develop branch to be tested by some nightly builds.


edhartnett:
Can you hold off on merge to develop?

I'm pretty sure I messed it up a bit. What I would like to do is get the current PRs merged, and the re-create develop. Hopefully should be possible to do that today.

I just did a merge to master which was correct, but I don't think the merge to develop was correct. So I think we need to re-create develop from master, or otherwise fix it. If you want to try and fix it, that would be great. I have no other outstanding PRs.

edhartnett:
I'm going to merge these changes to master, and then recreate the develop branch.

PR #347, opened 2016-12-17
update compiler and netcdf libraries

jedwards4b:
Fixes issues with tests failing in cdash. netcdf 4.3.3.1 has known problems, updated to netcdf 4.4.1
also updated gnu to 6.1.0 and pnetcdf to 1.7.0.

Already merged to develop.

PR #358, opened 2016-12-20
Adding tests for pio inquire functions

jayeshkrishna:
Adding a test for some pio (variable, dim, att) inquire functions.
More tests are required for better coverage, but this is a
start.

### DIFF
--- a/tests/general/CMakeLists.txt
+++ b/tests/general/CMakeLists.txt
@@ -11,6 +11,7 @@ SET(GENERATED_SRCS pio_file_simple_tests.F90
   ncdf_simple_tests.F90
   ncdf_get_put.F90
   ncdf_fail.F90
+  ncdf_inq.F90
   pio_decomp_tests.F90
   pio_decomp_tests_1d.F90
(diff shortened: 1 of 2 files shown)


jayeshkrishna:
In most cases the return values are checked for correctness (PIO_TF_PASSERT*) but we could add more checks like making sure that all ranks get the same var/dim ids.

jayeshkrishna:
I will soon add a feature to the testing framework to check if all tasks have the same value for a variable. Then it would be easy to add these checks.

jedwards4b:
I don't see this in develop yet.

jayeshkrishna:
I am holding off my merge to develop because when I merge my branch to develop it also brings in update_yellowstone_gnu-2 (part of PR #347). It looks like PR #347 was not merged to develop.

jayeshkrishna:
Just verified that the change in PR #347 is on develop. Will merge this branch soon.

jayeshkrishna:
This PR should be ready to merge to master (passes on cdash)

PR #369, opened 2016-12-22
Ejh error handling 2

edhartnett:
More of #225.


src/clib/pio_darray_int.c
```diff
@@ -1419,13 +1425,13 @@ int flush_output_buffer(file_desc_t *file, bool force, PIO_Offset addsize)
     PIO_Offset usage = 0;
 
     /* Check inputs. */
-    if (!file)
-        return PIO_EINVAL;
+    pioassert(file, "invalid input", __FILE__, __LINE__);
 
-    /* Find out the buffer usage. */
+    /* Find out the buffer usage. If I check the turn code, some tests fail. ??? */
```
jedwards4b: Are we calling it on some tasks or in some cases with no buffer?

edhartnett: Apparently.

PR #378, opened 2016-12-27
fixed nelems

edhartnett:
We need a test, but here is the fix.

I will merge with develop, then do a test tomorrow...


jedwards4b:
A unit test is good, but fyi I tested with cesm and it solved that issue.  Thanks

Issue #379, opened 2016-12-28
Need tests for put/get var/var1/vara/vars functions


edhartnett:
There are currently no tests for these crucial functions.

edhartnett:
Merged some tests for these today. Still many more to go before this ticket gets closed.

Would like to test:
* All different ways of getting/putting data (var1, vara, vars) EXCEPT varm, which is not supported in async.
* All different types should be used, including netCDF-4 enhanced types, on builds for which _NETCDF4 is true.
* At least a few combinations of index, count, and stride.
* For record and non-record variables.
* Data should be checked before file is closed (after sync for pnetcdf) and with a re-opened file.

edhartnett:
Some tests have been added. More to come.

PR #381, opened 2016-12-29
started to add tests for put/get

edhartnett:
Added a section to do a put and then a get.

But the get does not work. Not yet clear why so it is commented out while I work on it. But I want to merge at least the one put test I've added, to get the ball rolling.

This PR contains test code changes only. No library code is changed.

This is part of #379. Also part of #205.

### DIFF
--- a/tests/cunit/test_pioc.c
+++ b/tests/cunit/test_pioc.c
@@ -446,6 +446,82 @@ int test_names(int iosysid, int num_flavors, int *flavor, int my_rank,
     return PIO_NOERR;
 }
 
+/* Test data read/write operations.
+ *
+ * @param iosysid the iosystem ID that will be used for the test.
+ * @param num_flavors the number of different IO types that will be tested.
+ * @param flavor an array of the valid IO types.
+ * @param my_rank 0-based rank of task.
+ * @returns 0 for success, error code otherwise.
+ */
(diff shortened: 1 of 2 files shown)


tests/cunit/test_pioc.c
```diff
@@ -446,6 +446,82 @@ int test_names(int iosysid, int num_flavors, int *flavor, int my_rank,
...
+        if ((ret = PIOc_enddef(ncid)))
+            ERR(ret);
+
+        /* Write some data. */
+        PIO_Offset start[NDIM] = {0, 0, 0};
+        PIO_Offset count[NDIM] = {1, 1, 1};
+        float data = 42.42;
+        if ((ret = PIOc_put_vara_float(ncid, varid, start, count, &data)))
+            ERR(ret);
+
```
jedwards4b: Closing the file here and reopening it to read will solve the problem.

edhartnett: I just discovered that. Why? Seems like a bug in pnetcdf. The other netcdf variants handle it fine, only pnetcdf fails.

jedwards4b: Actually calling PIOc_sync(ncid) will solve the problem too.

jedwards4b: In pnetcdf small writes are stored in memory until a sync or close file - or the memory buffer is exhausted - this is a performance feature.

PR #382, opened 2016-12-29
this cannot be collective

### LINKED ISSUE #118: Should there be one global cn_buffer_pool, or should each iosysid get a CN buffer?
Right now there is one cn_buffer pool for all open iosystems. Is that correct?

jedwards4b:
Since we are calling this on the last iosystem on each task and these iosystems do not necessarily reflect a common communicator we cannot use collective calls.

Fixes #118


jedwards4b:
merged to develop

PR #385, opened 2016-12-29
Changes to init async API

### LINKED ISSUE #363: Change InitAsync to return copies of IO and component communicators
THe calling program also needs these communicators, so make copies of them for the user.

This is necessary to solve #349 and #348.

edhartnett:
Fixes #363.


edhartnett:
Because calling programs using async need to be able to call things on the computation communicator, but not the IO communicator.

That is, after async is set up, MPI_COMM_WORLD can't be used, because the IO tasks are busy listening for messages.

PR #390, opened 2016-12-31
added some netcdf-4 types to putget tests

edhartnett:
Part of #379. Still plenty of work to do here.

I will merge these changes to develop for testing.


src/clib/pio.h
```diff
@@ -438,6 +438,12 @@ enum PIO_ERROR_HANDLERS
 #define PIO_FLOAT  NC_FLOAT
 #define PIO_REAL   NC_FLOAT
 #define PIO_DOUBLE NC_DOUBLE
+#define PIO_UBYTE  NC_UBYTE
```
jedwards4b: Are all of this defined in pnetcdf as well?

edhartnett: No, they are not defined in pnetcdf.

edhartnett: Not only that, but it seems the pnetcdf supports all the netcdf-4 types except string. Very nice.

PR #406, opened 2017-01-04
Laramie update

jedwards4b:
update to pnetcdf/1.8.0
I will merge to develop


edhartnett:
Looking at the commits, it looks like you branched off develop instead of master?

PR #408, opened 2017-01-05
fixed memory leak

### LINKED ISSUE #404: IO comm created by PIOc_Init_Async() not freed
In PIOc_Init_Async(), a shared IO comm is created:
```
/* There is one shared IO comm. Create it. */
if ((ret = MPI_Comm_create(world, io_group, &io_comm)))
    return check_mpi(NULL, ret, __FILE__, __LINE__);
```
Only a copy (dup) of the IO comm is saved in my_iosys (iosystem_desc_t struct):
```
/* If this is the IO component, remember the
 * comm. Otherwise make a copy of the comm for each
 * component. */
if (in_io)
    if (cmp)
    {
        LOG((3, "making a dup of io_comm = %d io_rank = %d", io_comm, io_rank));
        if ((ret = MPI_Comm_dup(io_comm, &my_iosys->io_comm)))
            return check_mpi(NULL, ret, __FILE__, __LINE__);
        ...
    }
```
Two minor issues with the code above:
1) The comment seems to be out of date (probably for old code)
2) It is better to add {} to the first if statement, to achieve better code readability

In PIOc_finalize() where we will free MPI resources, only the copy of the IO comm is freed
```
if (ios->io_comm != MPI_COMM_NULL)
    MPI_Comm_free(&ios->io_comm);
```
All of the async tests have this leak. We need to save the original io_comm in some variable, such that we can call MPI_Comm_free on it later in PIOc_finalize().

edhartnett:
Fixes #404.

### DIFF
--- a/src/clib/pioc.c
+++ b/src/clib/pioc.c
@@ -1309,6 +1309,10 @@ int PIOc_Init_Async(MPI_Comm world, int num_io_procs, int *io_proc_list,
     if (!io_proc_list)
         free(my_io_proc_list);
 
+    if (in_io)
+        if ((mpierr = MPI_Comm_free(&io_comm)))
+            return check_mpi(NULL, ret, __FILE__, __LINE__);
+    
     if (!proc_list)
     {
         for (int cmp = 0; cmp < component_count + 1; cmp++)


dqwu:
Looks good. I tested the latest develop branch, test_async_simple has no memory leaks. It also reduces leaks in test_async_3proc, test_async_4proc and test_pioc.

Could you please also update the comment in pioc.c, line 1228?
It was written for the old code shown below:
```
/* If this is the IO component, remember the
 * comm. Otherwise make a copy of the comm for each
 * component. */
if (in_io)
    if (!cmp)
    {
        LOG((3, "setting io_comm to %d", my_iosys->io_comm));
        io_comm = my_iosys->io_comm;
    }
    else
    {
        LOG((3, "making a dup of io_comm = %d", io_comm));		
        if ((ret = MPI_Comm_dup(io_comm, &my_iosys->io_comm)))
            return check_mpi(NULL, ret,__FILE__,__LINE__);
    }
```

PR #411, opened 2017-01-05
fix an error in knowing if you need to fill holes

### LINKED ISSUE #410: pioperf test issue
Trying to reduce some rather complex problems into small test cases.   This case uses the pioperf
executable in tests/performance on 4 tasks with 2 io tasks.  The namelist is
```
&pioperf
  decompfile= 'simple.4', 'simple.4_1hole'
  pio_typenames = 'pnetcdf'
  rearrangers = 2
  niotasks = 2
  nframes = 1
  nvars = 1
  varsize = 4
  unlimdimindof = .false.
/
```
the first decomp file works correctly, the second writes correctly but then fails on the open file
for read.   The two decomp files are attached (with a txt extension)
[simple.4.txt](https://github.com/NCAR/ParallelIO/files/688003/simple.4.txt)
[simple.4_1hole.txt](https://github.com/NCAR/ParallelIO/files/688004/simple.4_1hole.txt)

jedwards4b:
Fixes #410 
If any iotask has holegridsize > 0 then all must enter this code.

### DIFF
--- a/src/clib/pio_darray.c
+++ b/src/clib/pio_darray.c
@@ -112,6 +112,7 @@ int PIOc_write_darray_multi(int ncid, const int *vid, int ioid, int nvars, PIO_O
 
     /* Currently there are two rearrangers box=1 and subset=2. There
      * is never a case where rearranger==0. */
+    LOG((2, "pio_write_darray_multi nvars=%d arraylen=%ld \n",nvars, arraylen));
     if (iodesc->rearranger > 0)
     {
         if (rlen > 0)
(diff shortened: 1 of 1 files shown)


jedwards4b:
merged to develop

src/clib/pio_darray.c
```diff
@@ -186,8 +187,9 @@ int PIOc_write_darray_multi(int ncid, const int *vid, int ioid, int nvars, PIO_O
      * points. This is generally faster than the netcdf method of
      * filling the entire array with missing values before overwriting
      * those values later. */
-    if (iodesc->rearranger == PIO_REARR_SUBSET && iodesc->needsfill &&
-        iodesc->holegridsize > 0)
+    if (iodesc->rearranger == PIO_REARR_SUBSET)
+	LOG((2, "pio_write_darray_multi nvars=%d holegridsize=%ld %d\n",nvars, iodesc->holegridsize, iodesc->needsfill));
+    if (iodesc->rearranger == PIO_REARR_SUBSET && iodesc->needsfill)
```
jayeshkrishna: bget() assumes requested size > 0 (when not using malloc), so does this change cause issues on procs where holegridsize is 0 ?

jayeshkrishna: But as you noted all procs should be getting into the write call

PR #415, opened 2017-01-06
add PIO_DATATYPE_NULL with switch depending on MPI variety

jedwards4b:
We had a workaround for MPI_DATATYPE_NULL for OPEN_MPI in pio_spmd.c, it turns out that SGI's MPT also needs this workaround.  So I added a PIO_DATATYPE_NULL macro and moved the definition to pio_internal.h OPEN_MPI and MPT_VERSION are macros defined in the respective MPI libraries so no further changes to the build are required.

### DIFF
--- a/src/clib/pio_internal.h
+++ b/src/clib/pio_internal.h
@@ -23,6 +23,14 @@
 #define MPI_Offset long long
 #endif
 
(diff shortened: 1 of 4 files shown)


jedwards4b:
Pushed to develop for testing

PR #416, opened 2017-01-06
added way to save decomp file

### LINKED ISSUE #213: Needs to be a way to set PIO_Save_Decomps to save decompositions to file
There needs to be a way to set PIO_Save_Decomps. It used to be set by environment variable that is not compatible with the netCDF API. However, since it is useful, I would like to make it available to netCDF users.

We can make this available through the API in some way.

We could make the default a CMake option.

We could also automatically turn this on when the log level is set to non-zero. Would that be useful?

We can include backward support for the environment variable if that is required for current users. I would prefer not to, but it can be done if necessary. I can add a build option to turn off this environment var support and use that build option with netCDF.

edhartnett:
Fixes #213.

### DIFF
--- a/src/clib/pio.h
+++ b/src/clib/pio.h
@@ -104,8 +104,11 @@ typedef struct io_desc_t
     /** The ID of this io_desc_t. */
     int ioid;
 
-    /** Not used... */
-    /*int async_id;*/
+    /** The length of the decomposition map. */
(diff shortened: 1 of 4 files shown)


edhartnett:
Whoops! Those were commented out in debugging. Now restored.

edhartnett:

PR #460, opened 2017-01-09
 Check compmaster using MPI_ROOT

### LINKED ISSUE #418: Issues with iomaster and compmaster
In pio.h, struct iosystem_desc_t contains two members: iomaster and compmaster
```
/** Set to MPI_ROOT if this task is the master of IO communicator, 0
 * otherwise. */
int iomaster;

/** Set to MPI_ROOT if this task is the master of comp communicator, 0
 * otherwise. */
int compmaster;
```
According to the comments above, iomaster or compmaster will be set to MPI_ROOT or 0
There are many places in the code that they are tested against 0
```
if (ios->iomaster)
    rlen = iodesc->maxiobuflen;
...
/* Send the message to the message handler. */
if (ios->compmaster)
    mpierr = MPI_Send(&msg, 1, MPI_INT, ios->ioroot, 1, ios->union_comm);
```
However, in pioc.c, they will be set to either MPI_ROOT or MPI_PROC_NULL
```
/* Set to MPI_ROOT on master process, MPI_PROC_NULL on other
 * processes. */
int iomaster;
...
iomaster = !io_rank ? MPI_ROOT : MPI_PROC_NULL;
...
/* Set comp_rank 0 to be the compmaster. It will have
 * a setting of MPI_ROOT, all other tasks will have a
 * setting of MPI_PROC_NULL. */
my_iosys->compmaster = my_iosys->comp_rank ? MPI_PROC_NULL : MPI_ROOT;
```

If MPI_PROC_NULL is 0, there should be no issues. However, we will find out that MPI_PROC_NULL is actually defined as -1 in mpi.h
```
#define MPI_PROC_NULL   (-1)
#define MPI_ANY_SOURCE 	(-2)
#define MPI_ROOT        (-3)
```
Therefore, if iomaster or compmaster is either MPI_ROOT or MPI_PROC_NULL, they will always be nonzero when we test with if (ios->iomaster) or if (ios->compmaster). Due to this bug, all compute tasks in async tests will call MPI_Send which should only be called by the master task (MPI memory leaks observed).

[Two possible fixes]
a) iomaster and compmaster will be MPI_ROOT or MPI_PROC_NULL. Then we need to update the comments in pio.h, and update 100+ lines of code to if (ios->iomaster == MPI_ROOT) or if (ios->compmaster == MPI_ROOT)

b)  iomaster and compmaster will be MPI_ROOT or 0
This seems less work, as we need only change two lines:
```
my_iosys->compmaster = my_iosys->comp_rank ? 0 : MPI_ROOT;
iomaster = !io_rank ? MPI_ROOT : 0;
```
However, this will affect a lot of existing MPI calls where we use MPI_PROC_NULL as a dummy source or destination for communication, like the call below:
`MPI_Bcast(&file->pio_ncid, 1, MPI_INT, ios->compmaster, ios->intercomm);`
This fix will make some async tests hang there (timeout).

Therefore, we should go with a), although it needs more lines of code to be changed.

dqwu:
Fixes #418 (in addition to PR #419).

### DIFF
--- a/src/clib/pio_getput_int.c
+++ b/src/clib/pio_getput_int.c
@@ -540,7 +540,7 @@ int PIOc_put_vars_tc(int ncid, int varid, const PIO_Offset *start, const PIO_Off
         {
             int msg = PIO_MSG_PUT_VARS;
 
-            if(ios->compmaster)
+            if (ios->compmaster == MPI_ROOT)
                 mpierr = MPI_Send(&msg, 1, MPI_INT, ios->ioroot, 1, ios->union_comm);
 
             /* Send the function parameters and associated informaiton
--- a/src/clib/pio_put_nc.c
+++ b/src/clib/pio_put_nc.c
@@ -1139,8 +1139,8 @@ int PIOc_put_var(int ncid, int varid, const void *buf, PIO_Offset bufcount,
 
     msg = PIO_MSG_PUT_VAR;
 
-    if(ios->async_interface && ! ios->ioproc){
-        if(ios->compmaster)
+    if (ios->async_interface && ! ios->ioproc){
+        if (ios->compmaster == MPI_ROOT)
             mpierr = MPI_Send(&msg, 1,MPI_INT, ios->ioroot, 1, ios->union_comm);
         mpierr = MPI_Bcast(&ncid,1, MPI_INT, ios->compmaster, ios->intercomm);
     }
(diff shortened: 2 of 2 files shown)


jedwards4b:
Was this merged to develop?

edhartnett:
I believe it was, and has already passed through cdash testing.

@dqwu suggest you merge because my current PR has some minor conflicts with this. ;-) You probably want to get your changes in first.

dqwu:
The feature branch was merged to develop for testing. This PR can be merged to master now.

Issue #245, opened 2016-11-30
Fix memory leaks in C tests


edhartnett:
From Danqing:

> I also created a new Jenkins builder for valgrind check, currently it uses feature branch dqwu/valgrind_check
> https://jenkins-ci.mcs.anl.gov/job/ParallelIO_valgrind
> 
> Almost every tests have memory leaks. For the first one (1/74 Test: test_darray), I have looked at 
> definitely lost: 1,459 bytes in 36 blocks
> 
> In tests/unit/test_darray.c, there is a function check_file() that has memory leaks on MPI types (MPI_Type_commit is called, but MPI_Type_free is NOT)
> 
> In function main() we call PIOc_InitDecomp(), and we later call PIOc_freedecomp() that can invoke MPI_Type_free for cleaning.
> In check_file(), we call create_decomposition(), which calls PIOc_InitDecomp(), but there is no call to PIOc_freedecomp() at the end of this function for cleaning like main().
> Note: in check_file() we have a locally defined variable ioid that will have a different value as ioid defined in main()
> 
> After I apply a local fix, definitely lost bytes reduced from 1,459 bytes to 32 bytes.

edhartnett:
These are not actually in the test code, they are in the library. ;-)

edhartnett:
Are we clear of memory leaks now, or are there still some left?

dqwu:
For Jenkins builder ParallelIO_valgrind_develop, there are no memory leaks on build 46. However, on build 47 there are new leaks from pio_rearr_opts. On build 48, test_pioc_putget and pio_rearr_opts failed due to timeout. We need look at recent PRs that got merged to the develop.

jayeshkrishna:
@dqwu: Please create issues for mem leaks associated with pio rearranger opts and assign it to me.

dqwu:
On build 49, test_pioc_putget and pio_rearr_opts passed. Maybe the timeout issue on build 48 is a false positive caused by the load of build server.

jayeshkrishna:
Yes, the pio_rearr_opts is a relatively long test (~40s on my laptop) and must have timed out due to load on the server.

edhartnett:
I will close this ticket. New leaks should be fixed before merge, and if not, will be given their own issues.

PR #478, opened 2017-01-11
Fix memory leaks for test_pioc

dqwu:
Part of #245.

### DIFF
--- a/src/clib/pio_msg.c
+++ b/src/clib/pio_msg.c
@@ -135,18 +135,30 @@ int create_file_handler(iosystem_desc_t *ios)
         return pio_err(ios, NULL, PIO_ENOMEM, __FILE__, __LINE__);
     if ((mpierr = MPI_Bcast((void *)filename, len + 1, MPI_CHAR, 0,
                             ios->intercomm)))
+    {
+        free(filename);
         return check_mpi2(ios, NULL, mpierr, __FILE__, __LINE__);
+    }
     if ((mpierr = MPI_Bcast(&iotype, 1, MPI_INT, 0, ios->intercomm)))
+    {
+        free(filename);
         return check_mpi2(ios, NULL, mpierr, __FILE__, __LINE__);
+    }
     if ((mpierr = MPI_Bcast(&mode, 1, MPI_INT, 0, ios->intercomm)))
+    {
+        free(filename);
         return check_mpi2(ios, NULL, mpierr, __FILE__, __LINE__);
+    }
     LOG((1, "create_file_handler got parameters len = %d "
          "filename = %s iotype = %d mode = %d\n",
          len, filename, iotype, mode));
 
     /* Call the create file function. */
     if ((ret = PIOc_createfile(ios->iosysid, &ncid, &iotype, filename, mode)))
+    {
+        free(filename);
         return pio_err(ios, NULL, ret, __FILE__, __LINE__);
+    }
 
     /* Free resources. */
     free(filename);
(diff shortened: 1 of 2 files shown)


src/clib/pio_msg.c
```diff
@@ -378,11 +399,20 @@ int inq_att_handler(iosystem_desc_t *ios)
...
     if ((mpierr = MPI_Bcast(&xtype_present, 1, MPI_CHAR, 0, ios->intercomm)))
+    {
+        free(name);
         return check_mpi2(ios, NULL, mpierr, __FILE__, __LINE__);
+    }
     if ((mpierr = MPI_Bcast(&len_present, 1, MPI_CHAR, 0, ios->intercomm)))
+    {
+        free(name);
         return check_mpi2(ios, NULL, mpierr, __FILE__, __LINE__);
+    }
```
edhartnett: Too much duplication of code. Instead, change to:

if (!mpierr)
    mpierr = MPI_Bcast(...)

Then at the end of all the calls
if (mpierr)
{
   free(name);
   check_mpi2(...);
}

src/clib/pio_msg.c
```diff
@@ -531,28 +573,53 @@ int att_put_handler(iosystem_desc_t *ios)
...
          "atttype = %d attlen = %d typelen = %d",
          ncid, varid, namelen, name, iotype, atttype, attlen, typelen));
 
     /* Call the function to read the attribute. */
     if ((ret = PIOc_put_att(ncid, varid, name, atttype, attlen, op)))
+    {
+        free(name);
+        free(op);
         return pio_err(ios, NULL, ret, __FILE__, __LINE__);
+    }
```
edhartnett: Same comment. No need to add all these lines of code.

src/clib/pio_msg.c
```diff
@@ -1638,13 +1796,19 @@ int delete_att_handler(iosystem_desc_t *ios)
...
     LOG((2, "delete_att_handler namelen = %d name = %s ncid = %d varid = %d ",
          namelen, name, ncid, varid));
 
     /* Call the create file function. */
     if ((ret = PIOc_del_att(ncid, varid, name)))
+    {
+        free(name);
         return pio_err(ios, NULL, ret, __FILE__, __LINE__);
+    }
```
edhartnett: I suspect there is a better way to handle this issue than just adding free(name) everywhere. How about a sole exit point for errors where it is freed? Or how about we use a declared instead of mallocd name value? If declared at size PIO_MAX_NAME + 1 it will always be long enough and does not have to be freed everywhere.

edhartnett:
You have made a great many changes in pio_msg. Was the code there causing memory loss in tests?

src/clib/pio_msg.c
dqwu: The issue is that in current code, we have a lot of places where we would return earlier on an MPI error
```
if ((mpierr = MPI_Bcast(...))
    return check_mpi(NULL, mpierr, __FILE__, __LINE__);
```
If you do not like the above code, we can change to the new structure you suggested.

src/clib/pio_msg.c
dqwu: I also do not know the reason why we use heap based memory for a name value. If we all agree to use a stack based one, we can get rid of free(name) everywhere, and do not have to change the existing code like:
```
if ((mpierr = MPI_Bcast(...))
    return check_mpi2(NULL, mpierr, __FILE__, __LINE__);
```

dqwu:
The code has caused memory leaks in open_file_handler()
We have tests that PIOc_openfile() will  return an error code without freeing filename
```
/* Call the open file function. */
if ((ret = PIOc_openfile(ios->iosysid, &ncid, &iotype, filename, mode)))
    return pio_err(ios, NULL, ret, __FILE__, __LINE__);
```
I added free in more places that would have similar issues (though the existing tests do not return errors there so far).

We need to decide if we should like to use stacked based name values in pio_msg.c

src/clib/pio_msg.c
jedwards4b: I agree - lets move to char name[PIO_MAX_NAME+1];

dqwu: Thanks, I will updated the PR later with stack based name values.

dqwu:
Merged latest PR to develop.

src/clib/pio_msg.c
```diff
@@ -131,37 +131,21 @@ int create_file_handler(iosystem_desc_t *ios)
     if ((mpierr = MPI_Bcast(&len, 1, MPI_INT, 0, ios->intercomm)))
         return check_mpi2(ios, NULL, mpierr, __FILE__, __LINE__);
     LOG((1, "create_file_handler got parameter len = %d\n", len));
```
edhartnett: You don't have to put "\n" at the end of LOG statements. It is added automatically. When you have them, it adds unneeded blank lines in the log.

dqwu: This LOG line is in existing code. I will create another issue on cleaning up pio_msg.c

edhartnett:
I am going to merge this now. Everything has passed cdash testing, and I need to merge some other changes to pio_msg.c as well.

src/clib/pio_msg.c
edhartnett: That's because I converted some of my old printf statements to log statements when I added the logging system, but forgot to take out the \n. So that one's on me. ;-)

I have a PR up now with all the \n's removed.

Issue #438, opened 2017-01-07
Tests for functions in pio_smpd.c


edhartnett:
The code in pio_spmd.c is almost completely untested.

jayeshkrishna:
The pio runtime flow control options that I am working on (Issue #120) will bring in some tests for functions in this file,

edhartnett:
The pio_spmd.c file is being tested much more now, up fro <10% coverage to 60%, which is great progress.

The pio_swapm() function is now tested.

The only remaining function that is not tested is pio_fc_gatherv(). This function is a future optimization so is not currently used in the library.

edhartnett:
I will close this ticket since I think this is all the tests we are are planning on adding right now.

PR #479, opened 2017-01-11
Adding runtime flow control options to PIO2

jayeshkrishna:
Adding runtime flow control options to PIO2. Users can now pass
runtime flow control options to the library via the PIO_Init() call.

Fixes Issue #120

### DIFF
--- a/src/clib/pio.h
+++ b/src/clib/pio.h
@@ -98,6 +98,70 @@ typedef struct io_region
 } io_region;
 
 /**
+ * Note: The rearranger option values must match the definitions
+ * in the fortran interface
+ */
+/**
+ * Rearranger comm type
+ */
+enum PIO_REARR_COMM_TYPE
+{
+    /** Point to point */
+    PIO_REARR_COMM_P2P = (0),
+    /** Collective */
+    PIO_REARR_COMM_COLL
+};
+
+/**
+ * Rearranger comm flow control direction
+ */
+enum PIO_REARR_COMM_FC_DIR
+{
+    /** Comp procs to io procs and vice versa */
+    PIO_REARR_COMM_FC_2D_ENABLE = (0),
+    /** Comp procs to io procs only */
+    PIO_REARR_COMM_FC_1D_COMP2IO,
+    /** IO procs to comp procs only */
+    PIO_REARR_COMM_FC_1D_IO2COMP,
+    /** Disable flow control */
+    PIO_REARR_COMM_FC_2D_DISABLE
+};
+
+/**
+ * Rearranger comm flow control options
+ */
+#define PIO_REARR_COMM_UNLIMITED_PEND_REQ -1
+typedef struct rearr_comm_fc_opt{
+    /* FIXME: Is _Bool instead of bool required here? */
+    /** Enable handshake */
+    bool enable_hs;
+    /** Enable isends  - if false use blocking sends */
+    bool enable_isend;
+    /** Max pending requests
+      * (PIO_REARR_COMM_UNLIMITED_PEND_REQ => unlimited pend req)
+      */
+    /** This is the number of messages allowed to be in flight at one
+     * time. On some systems posting all messages at once creates a
+     * significant bottleneck in communications and throttling in this
+     * manner improves overall performance. */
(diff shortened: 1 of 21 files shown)


jayeshkrishna:
The development branch was rebased onto a recent version of master to make the merge process easier (removed conflicts with master)

jayeshkrishna:
The added test should also exercise functions in pio_spmd.c (See Issue #438)

src/clib/pio_spmd.c
```diff
@@ -288,10 +288,26 @@ int pio_swapm(void *sendbuf, int *sendcounts, int *sdispls, MPI_Datatype *sendty
...
+             * a major issue anymore, or is buggy. With PIO1 we have found that
+             * although the code correctly posts receives before the irsends,
+             * on some systems (software stacks) the code hangs. However the 
+             * code works fine with isends. The NO_MPI_RSEND macro should be 
+             * used to choose between mpi_irsends and mpi_isends - the default
+             * is still mpi_irsend
+             */ 
             if (isend)
+            {
+#ifndef NO_MPI_IRSEND
```
jedwards4b: can you swap these to remove the double negative?

src/clib/pio.h
```diff
@@ -595,7 +653,7 @@ extern "C" {
                             MPI_Comm io_comm, int *iosysidp);
     int PIOc_get_numiotasks(int iosysid, int *numiotasks);
     int PIOc_Init_Intracomm(MPI_Comm comp_comm, int num_iotasks, int stride, int base, int rearr,
-                            int *iosysidp);
+                            rearr_opt_t *rearr_opts, int *iosysidp);
```
jedwards4b: I think that Ed is going to argue that we do not want derived types as part of our PIOc interface.   
Can we maintain the derived type in fortran but list all 6 opts individually in the PIOc functions?

src/clib/pio_spmd.c
jayeshkrishna: ok, would changing it to,
```#ifdef USE_MPI_ISEND_FOR_FC```
work?

src/clib/pio.h
jayeshkrishna: Sure, we can do that.

src/clib/pio_spmd.c
jedwards4b: Yes that should be fine.  Thanks

src/clib/pio.h
jayeshkrishna: One issue however is that it would be harder (if the rearranger flow control options are explicitly listed) for users to ignore setting the rearranger options

jedwards4b: Yes I wonder if we should leave them out of the Init_Intracomm interface and have separate functions to set them - it seems like eventually we may want finer control of these variables and setting them in init functions limits that possibility.

jayeshkrishna: Yeah, I was thinking of adding a new function to set flow control parameters after init (after merging this change).
Ok, In that case I will add a new interface to set the flow control parameters (initialize it to defaults in pio_init())

jayeshkrishna: i,e, Get the flow control parameters out of pio_init()

edhartnett: I would prefer if you addressed Jim's issues before you merged to master.

jayeshkrishna: Yeah, this change is not going to develop/master until we address these issues

jayeshkrishna:
The user rearranger options have been removed from the C interface. A new function (C and Fortran), pio_set_rearr_opts(), has been added to set the rearranger options .
The branch was also rebased (making merge easier - there were conflicts with master)

jayeshkrishna:
@jedwards4b : Please take a look at this PR now (all the changes requested has been fixed).

jedwards4b:
I've merged this to develop.   Thanks

jayeshkrishna:
The cdash output shows no failures. So this should be ready to merge to master.

edhartnett:
Shall we merge these to master?

jayeshkrishna:
@jedwards4b : These changes are ready to merge to master. Do you want me to do it?

PR #493, opened 2017-01-13
Fix return statements in wrong places for netCDF-4/HDF5 iotype in ncdf_tests.F90

### LINKED ISSUE #492: Early return in ncdf_tests.F90 for netCDF-4/HDF5 iotypes
For netCDF-4/HDF5 iotypes, even when there are no test errors, the subroutine test_nc4 unexpectedly return early, see the code below (a comment is added to the return line):
```
! Should not have worked except for netCDF-4/HDF5 iotypes.
if (iotype .eq. PIO_iotype_netcdf4c .or. iotype .eq. PIO_iotype_netcdf4p) then
   if (ret_val .ne. PIO_NOERR) then
      err_msg = "Could not set chunk cache"
      call PIO_closefile(pio_file)
   endif
   if (chunk_cache_size_in .ne. chunk_cache_size .or. chunk_cache_nelems_in .ne. chunk_cache_nelems .or. &
        chunk_cache_preemption_in .ne. chunk_cache_preemption) then
      err_msg = "Incorrect chunk cache values!"
      call PIO_closefile(pio_file)
   endif
   return /* NOTE: early return here */
else if (iotype .eq. PIO_iotype_pnetcdf .and. ret_val .eq. PIO_NOERR) then
   err_msg = "Did not get expected error when trying to set chunk cache for pnetcdf file"
   call PIO_closefile(pio_file)
   return
else if (iotype .eq. PIO_iotype_netcdf .and. ret_val .eq. PIO_NOERR) then
   err_msg = "Did not get expected error when trying to set chunk cache for netcdf classic file"
   call PIO_closefile(pio_file)
   return
end if
```
This unexpected early return has two issues:
1) The following code are not tested for netCDF-4/HDF5 iotypes (code coverage issue)
2) There are observed memory leaks (for pio_unit_test), as PIO_freedecomp() at the end of test_nc4 will not be called for netCDF-4/HDF5 iotypes

The return statement should go to the two inner if statements (early return for errors only)

dqwu:
Fixes #492.

### DIFF
--- a/tests/unit/ncdf_tests.F90
+++ b/tests/unit/ncdf_tests.F90
@@ -349,13 +349,14 @@ Contains
        if (ret_val .ne. PIO_NOERR) then
           err_msg = "Could not set chunk cache"
           call PIO_closefile(pio_file)
+          return
        endif
        if (chunk_cache_size_in .ne. chunk_cache_size .or. chunk_cache_nelems_in .ne. chunk_cache_nelems .or. &
             chunk_cache_preemption_in .ne. chunk_cache_preemption) then
           err_msg = "Incorrect chunk cache values!"
           call PIO_closefile(pio_file)
+          return
        endif
-       return
     else if (iotype .eq. PIO_iotype_pnetcdf .and. ret_val .eq. PIO_NOERR) then
        err_msg = "Did not get expected error when trying to set chunk cache for pnetcdf file"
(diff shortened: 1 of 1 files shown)


edhartnett:
Can this be merged?

dqwu:
I think so, already tested on develop with reduced memory leaks.

PR #516, opened 2017-01-19
fixed issues in pio_msg.c

### LINKED ISSUE #512: Take name length asserts out of pio.msg.c
In pio_msg.c there is now code like this in various places:

```
    /* Get the parameters for this function that the he comp master
     * task is broadcasting. */
    if ((mpierr = MPI_Bcast(&len, 1, MPI_INT, 0, ios->intercomm)))
        return check_mpi2(ios, NULL, mpierr, __FILE__, __LINE__);
    LOG((1, "create_file_handler got parameter len = %d", len));
    pioassert(len <= PIO_MAX_NAME, "len > PIO_MAX_NAME", __FILE__, __LINE__);
    if ((mpierr = MPI_Bcast((void *)filename, len + 1, MPI_CHAR, 0,
                            ios->intercomm)))
        return check_mpi2(ios, NULL, mpierr, __FILE__, __LINE__);

```

The assert for the length of the name is not needed. This needs to be checked in the calling function, at the beginning of the function, and that's where it is already checked:

From PIOc_createfile():

```
    /* User must provide valid input for these parameters. */
    if (!ncidp || !iotype || !filename || strlen(filename) > NC_MAX_NAME)
        return pio_err(ios, NULL, PIO_EINVAL, __FILE__, __LINE__);

```

So the checks should be taken out of pio_msg.c, after confirming that the length of the name is checked in all the functions that end up calling the pio_msg.c functions.

### LINKED ISSUE #514: filename not guarenteed to be < PIO_MAX_NAME in open_file_handler()/create_file_handler() in pio_msg.c
From open_file_handler():

```
    char filename[PIO_MAX_NAME + 1];
    int mode;
    int mpierr;
    int ret;

    LOG((1, "open_file_handler comproot = %d", ios->comproot));
    assert(ios);

    /* Get the parameters for this function that the he comp master
     * task is broadcasting. */
    if ((mpierr = MPI_Bcast(&len, 1, MPI_INT, 0, ios->intercomm)))
        return check_mpi2(ios, NULL, mpierr, __FILE__, __LINE__);
    LOG((2, "open_file_handler got parameter len = %d", len));
    if ((mpierr = MPI_Bcast((void *)filename, len + 1, MPI_CHAR, 0,
                            ios->intercomm)))
        return check_mpi2(ios, NULL, mpierr, __FILE__, __LINE__);

```

Unlike almost all the other strings we pass through the API, the filename is not limited to NC_MAX_NAME. So we must allocate memory in this case.

A similar situation exists in create handler and rename handler.

edhartnett:
Fixes #514.
Fixes #512.

Some minor fixes for pio_msg.c.

### DIFF
--- a/src/clib/pio_msg.c
+++ b/src/clib/pio_msg.c
@@ -118,7 +118,6 @@ int create_file_handler(iosystem_desc_t *ios)
     int ncid;
     int len;
     int iotype;
-    char filename[PIO_MAX_NAME + 1];
     int mode;
     int mpierr;
     int ret;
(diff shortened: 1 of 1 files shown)


dqwu:
For PIO, does CMake have a requirement on the supported C compiler versions? I know that PETSc strictly support very old C compilers, so it will not use features from new compilers.

In this PR:
int len;
... /* get the value of len */    
char filename[len + 1];
This is supported by GCC since a certain version, but not by some old C compilers.

It seems that some C++ style code are already used in PIO. Another example is code like this (define a local variable within the for statement)
for (int d = 0; d < DIM_LEN; d++)

jayeshkrishna:
We expect C99 for sure

dqwu:
Does C99 support definition of an array whose size is a variable?

edhartnett:
We require C99, and that is a supported feature, so I think we're OK with it.

PR #521, opened 2017-01-19
Fix mem leak in rearr opts test

### LINKED ISSUE #520: Mem leak in pio_rearr_opts test
The test, pio_rearr_opts.F90.in, does not free all the MPI comms that it creates/dups (dup inside
loop but the free is outside the loop - looks like error during code refactoring).

@dqwu : Mentioned this bug in #245

jayeshkrishna:
The communicator used was dup'ed every time (to be used in
PIO init) inside the loop but only freed once (outside the
loop). Moving the free for the dup comm inside the loop.

Fixes #520

### DIFF
--- a/tests/general/pio_rearr_opts.F90.in
+++ b/tests/general/pio_rearr_opts.F90.in
@@ -451,9 +451,9 @@ PIO_TF_AUTO_TEST_SUB_BEGIN write_with_rearr_opts
     end do ! cur_rearr
 
     call PIO_deletefile(pio_tf_iosystem_, trim(tgv_fname))
-  end do ! iotypes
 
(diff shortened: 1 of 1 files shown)


jayeshkrishna:
@dqwu : Can you verify that this change fixes the mem leak in pio_rearr_opts test?

dqwu:
I tested this PR on my local machine and the memory leak is gone. You can merge it to develop for the build servers.

edhartnett:
Can we merge this?

dqwu:
This can be merged to master. Also, there are new memory leaks in master now, from test_iosystem3 and test_decomps.

PR #532, opened 2017-01-20
Update new case test_decomps

### LINKED ISSUE #531: Issues with test_decomps
In test_decomps.c, the arguments gdims and map passed to PIOc_readmap() are defined as arrays:
```
int gdims[NDIM];
PIO_Offset map[16];
```
This is incorrect, we should use pointers here. 
Note, C array names are constants, therefore they can not be changed by PIOc_readmap().

Issue 1:
Bad values observed
gdims[0] = 147042288
map[0] = 147042368
These values are addresses of allocated memory that PIOc_readmap() tries to assigns to gdims and map. The constant array names cannot be changed, instead, it is the first array element that got assigned. 

Issue 2:
Memory leaks. PIOc_readmap() allocated some memory internally and tried to assign the starting addresses to gdims and map (assigned to the first elements as described above). The memory should be freed later outside PIOc_readmap().

[How to fix]
Use pointers for gdims and map in this test, and free the allocated memory with them.

dqwu:
Fixes #531 
Will merge to develop for testing.

### DIFF
--- a/tests/cunit/test_decomps.c
+++ b/tests/cunit/test_decomps.c
@@ -54,9 +54,9 @@ int main(int argc, char **argv)
     int bad_slice_dimlen[2];    /* Invalid values. */
     int ioid;                   /* The decomposition ID. */
     int ndims;
-    int gdims[NDIM];
+    int *gdims = NULL;
     PIO_Offset fmaplen;
-    PIO_Offset map[16];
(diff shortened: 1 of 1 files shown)


edhartnett:
I'm going to merge this now, so I can add some more tests to test_decomps.c.

PR #562, opened 2017-01-31
Fix cunit test timeout issue on Valgrind builds

### LINKED ISSUE #561: Increase cunit test timeout for Valgrind builds
Valgrind check for test_pioc_putget is relatively long (~30s on my laptop). Due to the load on MCS build servers that cunit test has timed out a few times.

Currently, the default test timeout for a cunit test is 60 seconds. If the option PIO_VALGRIND_CHECK is set, we can set that value to 120 seconds.

dqwu:
Fixes #561.


dqwu:
@edhartnett 
Merged to develop for testing, and ParallelIO_valgrind_develop build passed (62.47 sec for test_pioc_putget). We can merge it to master now.

PR #569, opened 2017-02-03
fix issue in openfile_retry

jedwards4b:
The retry portion of openfile_retry was not working because file->do_io needed to be reset to the 
correct serial netcdf value.     There are several whitespace changes in this PR indicating that someone is still committing files with white space additions.

Will merge to develop for testing.

### DIFF
--- a/src/clib/pio_file.c
+++ b/src/clib/pio_file.c
@@ -498,7 +498,7 @@ int PIOc_sync(int ncid)
     if (file->mode & PIO_WRITE)
     {
         LOG((3, "PIOc_sync checking buffers"));
-        
+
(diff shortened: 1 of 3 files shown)


edhartnett:
Can you add a test for this too?

jedwards4b:
It appears this change has broken some of the tests, I'm looking into it now and will also add a test for the original problem.   Feel free to back out this change on develop if it is hindering your progress.

jedwards4b:
The problem was that the internal mode modifiers for netcdf4p and netcdf4c left the mode argument in a netcdf4 only value and so the classic netcdf call was failing.   I've updated the PR and pushed this correction to develop.

PR #603, opened 2017-02-07
Set log level when setting debug level

### LINKED ISSUE #576: Set log level in Fortran when debug level is set
Make sure that we set the log level when the debug level is set.
Currently the code just sets an internal debug level for the fortran
interface but does not set the logging level.

jayeshkrishna:
Making sure that we set the log level in the C library when user
sets the debug level via the Fortran interface.

Fixes #576


jayeshkrishna:
Yes, I thought about it. We still have some code in the fortran interface that might need logging/debugging (debugging turned on).
We could create an issue for cleaning up the fortran interface that can tackle (look/discuss if it is still needed) old code like this.

jayeshkrishna:
The cdash output looks good, merging to master

Issue #543, opened 2017-01-24
Fill value handling in  PIOc_write_darray_multi() wrong for some types


edhartnett:
The code only handles fill values properly for types that are 4 or 8 bytes long. So shorts and chars will not he handled correctly.

Issue #451, opened 2017-01-07
Test for PIOc_write_darray() without a fill value


edhartnett:
Test calling this function without providing a fill value parameter.

Issue #83, opened 2016-06-21
have PIOc_write_darray return error if given an incorrect fill value parameter


edhartnett:
It's not clear why PIOc_write_darray takes a fill value parameter.

Shouldn't it just use whatever fill value is set for that variable?

Seems like we risk users passing an incorrect value, and then thinking they are getting a fill value, when in fact they are just polluting the data.

So I wonder why the darray functions take a fill value?

edhartnett:
Seems like there is some buggy code here. If the user does not provide a fill value, the default fill value for the type is used. This ignores any fill value that the user has already set.

So if the user sets a fill value of -9999 and then call PIOc_write_darray() without providing a fill value argument, the default fill value will be used for those writes.

A similar error condition will occur if the user sets a fill value, and then provides a different fill value to PIOc_darray().

edhartnett:
I think we are going to want to eliminate the fill value parameter from the function call.

I think what we want to do instead is look up the netCDF fill value. (Attribute _FillValue). If that is not there, use the default fill value for the type.

jayeshkrishna:
Are we changing the API here?

edhartnett:
We may, but not necessarily the Fortran API.

jayeshkrishna:
I thought the consensus was not to change the API

edhartnett:
The C API may no longer be changed without good cause. In this case there is a bug in the actual API. We should not be asking the user to pass a new fill value. We should use the default, or the fill_value set in the file. If the user wants to set a fill value, he or she should do so in the standard way.

If we need to maintain backwards C compatibility we can add to the API, instead of changing it, but I think in this case the number of C API users is still low and so we can make this change. 

Areas of the C API that will be much more resistant to change are those functions that model netCDF functions, and use a copy of their parameters. However this function is a PIO-only function.

At the next meeting I will explain in more detail the policy for C API changes to anyone who is interested.

The Fortran API is a different story. I also recommend that you don't change it without excellent reason, but I have no requirements there - I'm not using that for the netCDF integration.

jayeshkrishna:
Bugs (in the code) should be fixed. But IMO the current capabilities of PIO should not be changed.

edhartnett:
The problem code is in pio_darray.c:

    ```
/* If we need a fill value, get it. */
    if (iodesc->needsfill)
    {
	/* If the user passed a fill value, use that, otherwise use
	 * the default fill value of the netCDF type. Copy the fill
[...]

PR #630, opened 2017-02-09
got var_inq_fill/var_def_fill working for classic files

### LINKED ISSUE #572: Change PIOc_inq_var_fill()/PIOc_def_var_fill() to work with classic netCDF
Right now these functions don't work for classic netCDF.

This is because classic netCDF does not provide the nc_def_var_fill()/nc_inq_var_fill() functions. This surprises me and I have submitted an issue on the netCDF project asking whether that is intentional. Seems like these functions should also be available in the classic library.

Since it's just setting/reading the _FillValue attribute, we can easily handle netCDF classic, and so provide this function for all iotypes. This would be handy for us internally as well.

edhartnett:
For some reason netcdf classic does not support nc_inq_var_fill() and nc_def_var_fill(). Not sure why, and I've asked.

Anyway, it would be tremendously helpful if we can use PIOc_def_var_fill()/PIOc_inq_var_fill() functions in other places, for example, in the darray code (#83, #543, #451), and count on them working for all iotypes. Now they do.

Fixes #572.


jayeshkrishna:
Was this PR reviewed?

edhartnett:
Not yet, but feel free to do so.

However the changes are pretty non-controversial, so I don't think anyone will have a problem with them.

This is part of on-going fill work that Jim is aware of. I assumed he missed this PR when he approved the PRs of the afternoon.

jayeshkrishna:
But the PR was merged before it was reviewed. If the reviewer ( @jedwards4b ) is aware+ok of the change I am ok with it too.

Issue #627, opened 2017-02-09
Add functions to write/read decomp files in netCDF format (instead of text)


edhartnett:
NetCDF files are more self-documenting and convenient, and it helps us "eat our own dogfood" if we use PIO to write/read the decomp files.

Here's what I have right now as the output. This is for a very simple case, a 1-D array of 4 elements, distributed over 4 tasks. Each task gets one element. (This is from test_pioc.c. More complex examples will follow.)

```
netcdf nc_decomp_test_pioc_rank_0_async_0 {
dimensions:
	dims = 1 ;
	task = 4 ;
	map_element = 1 ;
variables:
	int global_size(dims) ;
	int maplen(task) ;
	int map(task, map_element) ;

// global attributes:
		:version = 2017 ;
		:max_maplen = 1 ;
		:title = "Very Simple Test Decompositon" ;
		:history = "Added to PIO automatic testing by Ed in February 2017." ;
		:source = "Decomposition file produced by PIO library." ;
data:

 global_size = 4 ;

 maplen = 1, 1, 1, 1 ;

 map =
  0,
  1,
  2,
  3 ;
}


```

Some things to note:
* This is netCDF classic, which means I have to use 32-bit ints for the map values. By using netCDF-4 I could use 64-bit ints.
* It is zero-based, in accordance with C and netCDF custom. The C API will take 0-based data. We can still used 1-based assignments in Fortran, but the data file represents the C output and that's what ncdump will show. We will use fill-value where zero is used in the text decomp files.

jedwards4b:
so map_element is maxval(maplen)?  This should work fine - I think that for some of the largest maps we may need to use 64-bit ints, but we should be able to make that transition without user intervention or awareness shouldn't we?

edhartnett:
The problem with going to 64-bits is that we then must require netCDF-4, right?

There is no easy way to store 64-bit ints in netCDF classic.

edhartnett:
Here's what I have now...

```
netcdf nc_decomp_test_pioc_rank_0_async_0 {
dimensions:
	dims = 1 ;
	task = 4 ;
	map_element = 1 ;
variables:
	int global_size(dims) ;
	int maplen(task) ;
	int map(task, map_element) ;

// global attributes:
		:version = "2017" ;
		:max_maplen = 1 ;
		:title = "Very Simple Test Decompositon" ;
		:history = "Added to PIO automatic testing by Ed in February 2017." ;
[...]

Issue #417, opened 2017-01-06
need tests for saving and reading decomp files


edhartnett:
Including async cases.

PR #641, opened 2017-02-14
Write netCDF decomposition files, internal function to read them

edhartnett:
Part of #627. Also part of #417.

This is mostly new code. It won't impact any existing functionality.

Now have writing of decomp files to netCDF.

Also have an internal function that reads the netCDF decomp files.

Still need a public interface for reading, still need reading and writing to work on async, and still need more tests. ;-) But this code is tested.


src/clib/pioc_support.c
```diff
@@ -876,6 +877,453 @@ int PIOc_readmap_from_f90(const char *file, int *ndims, int **gdims, PIO_Offset
...
+
+    /* Write a source attribute. */
+    char source[] = "Decomposition file produced by PIO library.";
+    if ((ret = PIOc_put_att_text(ncid, NC_GLOBAL, DECOMP_SOURCE_ATT_NAME,
+                                 strlen(source) + 1, source)))
+        return pio_err(ios, NULL, ret, __FILE__, __LINE__);
+
+    /* Write an attribute with array ordering (C or Fortran). */
+    char c_order_str[] = DECOMP_C_ORDER_STR;
+    char fortran_order_str[] = DECOMP_FORTRAN_ORDER_STR;
```
jedwards4b: Maybe I didn't explain very well - I don't think there is a distinction here.

edhartnett: There is a distinction because if you write it out in C and read it in Fortran things may get confused.

(I think. Am I wrong about that?)

Issue #646, opened 2017-02-14
pair() function may be incorrect


edhartnett:
In pio_spmd.c we have this function:

```
/**
 * Given integers p and k between 0 and np-1,
 * if (p+1)^k <= np-1 then return (p+1)^k else -1.
 *
 * @param np
 * @param p integer between 0 and np - 1.
 * @param k integer between 0 and np - 1.
 * @returns (p + 1) ^ k else -1.
 */
int pair(int np, int p, int k)
{
    int q = (p + 1) ^ k ;
    int pair = (q > np - 1) ? -1 : q;
    return pair;
}

```
I didn't notice this until I tested it, but of course the ^ does not work as exponentiation in C, it is actually the bitwise XOR function, which I suspect is not what is intended here. Should pair be using the pow function instead?

jayeshkrishna:
There are several bugs in the pio_swapm() function. For now, please don't create any more issues related to functions in pio_spmd.c - we can do that once I finish the cleanup (once #637 is done).

jayeshkrishna:
xor is what is required here though.

edhartnett:
OK, very good. Suggest you mention the xor in the documentation for this function.

jayeshkrishna:
Although the code in pair() function can be refactored, there are no bugs in the function. Closing this issue.

Issue #645, opened 2017-02-14
is documentation of ceil2() function in pio_spmd.c correct?


edhartnett:
When writing a test for the ceil2() function, I found that it's documentation is wrong. It says:

> Returns the smallest power of 2 greater than i.

But it should be

> greater than or equal to i.

Mistake in code? Or mistake in documentation?

This function is used only one place, in pio_swapm().

jayeshkrishna:
Mistake in doc (as of the current code).
I am currently working on this code right now (and fixing docs as needed). You can assign this to me.

jayeshkrishna:
Doc added in 5fbe4c235f7dc4391324a6ed49f6a4bf194e12da

edhartnett:
OK, sounds good.

PR #647, opened 2017-02-14
added tests for ceil2() and pair() function in pio_spmd.c

### LINKED ISSUE #636: Test for ceil2() and pair() functions in pio_spmd.c
These functions are currently untested.

edhartnett:
Fixes #636. Part of #205.

Generated issues #645 and #646, but we can merge these working tests, which test the actual functionality of these functions. If we find that needs to change, the tests can change with it.

### DIFF
--- a/tests/cunit/test_spmd.c
+++ b/tests/cunit/test_spmd.c
@@ -352,6 +352,32 @@ int test_compare_offsets()
     return 0;
 }
 
+/* Test the ceil2() and pair() functions. */
+int test_ceil2_pair()
+{
+    /* Test the ceil2() function. */
+    if (ceil2(1) != 1)
+        return ERR_WRONG;
+    if (ceil2(-100) != 1)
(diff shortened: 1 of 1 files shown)


jayeshkrishna:
Do these tests pass?

jayeshkrishna:
And a question for everyone ( @jedwards4b / @edhartnett / @Katetc / @dqwu ) do we merge to develop before or after the review (refreshing my memory on the process)?

edhartnett:
These tests pass or I would not have merged them to develop. ;-)

edhartnett:
@jayeshkrishna although you will be adding tests, I would like to keep these tests as well.

My understanding is that your tests will be in fortran. I would also like to have C tests covering all the C code. So although something may be tested in fortran, I would still like to keep C tests.

jayeshkrishna:
I won't be removing any tests.

PR #651, opened 2017-02-15
Adding more 1d Fortran tests

jayeshkrishna:
Adding more 1d fortran tests,
* More combinations that force data rearrangement during read and write tested
* Testing reusing iodescs across reads (and reads+writes)

### DIFF
--- a/tests/general/pio_decomp_tests.F90.in
+++ b/tests/general/pio_decomp_tests.F90.in
@@ -99,7 +99,7 @@ PIO_TF_AUTO_TEST_SUB_BEGIN nc_write_1d_reuse_decomp
   character(len=PIO_TF_MAX_STR_LEN) :: filename1, filename2
   type(io_desc_t) :: iodesc
   integer, dimension(VEC_LOCAL_SZ) :: compdof, compdof_rel_disps
-  PIO_TF_FC_DATA_TYPE, dimension(VEC_LOCAL_SZ) :: buf
+  PIO_TF_FC_DATA_TYPE, dimension(VEC_LOCAL_SZ) :: buf, rbuf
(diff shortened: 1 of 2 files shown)


edhartnett:
Does this test check all different data types?

It looks to me like the malloc_iodesc function incorrectly sets the basetype for types < 4 bytes in size (i.e. shorts and chars). I believe that the only result is that later, too much memory is allocated to hold the data.

jayeshkrishna:
This test should test INTEGER, REAL (float/double) types. The fortran tests currently don't test other data types (types supported in netcdf that were added to pio)

jayeshkrishna:
Merged to develop - b4f89218ce8b5cddfe92f4119dbc9fa2a971f8b5

edhartnett:
Would you like me to merge this to master, or are you still working on these tests? Thanks!

jedwards4b:
Do not merge yet,  this is failing with multiple compiler errors on nag which has a hard line length limit of 132 characters.   It appears that the preprocessor is putting a comment at the end of each line of the F90 file with the line number from the original source.  This is unnecessary - if you use the cpp line control information you will get something that compilers can use to reference the correct source line.   
https://gcc.gnu.org/onlinedocs/cpp/Line-Control.html#Line-Control

jayeshkrishna:
I will try to see if I can get the line length less than 132 characters.
I have always had issues debugging code that has line control information in them. I think the testing framework has option to use line control if needed.

jayeshkrishna:
The line lengths should now be under 132 chars (I verified manually) and I have re-merged the change to develop.

jayeshkrishna:
The cdash results look good. Merging it to master now.

PR #656, opened 2017-02-16
Now read netCDF decomp file and InitDecomp() with values

### LINKED ISSUE #627: Add functions to write/read decomp files in netCDF format (instead of text)

### LINKED ISSUE #654: Test PIOc_write_nc_decomp() and related functions with bad inputs

edhartnett:
Added function PIOc_read_nc_decomp() to read netCDF decomp file and initialized iodesc. Plus tests. Fixes #627.

Also more tests for PIOc_write_nc_decomp(). Fixes #654.

### DIFF
--- a/src/clib/pioc_support.c
+++ b/src/clib/pioc_support.c
@@ -902,12 +902,16 @@ int PIOc_write_nc_decomp(const char *filename, int iosysid, int ioid, MPI_Comm c
     int mpierr;
     int ret;
 
-    LOG((1, "PIOc_write_nc_decomp filename = %s iosysid = %d ioid = %d", filename,
-         iosysid, ioid));
-
     /* Get the IO system info. */
     if (!(ios = pio_get_iosystem_from_id(iosysid)))
         return pio_err(NULL, NULL, PIO_EBADID, __FILE__, __LINE__);
+
+    /* Check inputs. */
+    if (!filename)
+        return pio_err(NULL, NULL, PIO_EINVAL, __FILE__, __LINE__);
+
+    LOG((1, "PIOc_write_nc_decomp filename = %s iosysid = %d ioid = %d", filename,
+         iosysid, ioid));
 
     /* Get the IO desc, which describes the decomposition. */
     if (!(iodesc = pio_get_iodesc_from_id(ioid)))
@@ -934,7 +938,7 @@ int PIOc_write_nc_decomp(const char *filename, int iosysid, int ioid, MPI_Comm c
 
     /* Find the max maxplen. */
     if ((mpierr = MPI_Allreduce(&iodesc->maplen, &max_maplen, 1, MPI_INT, MPI_MAX, comm)))
-        return check_mpi2(ios, NULL, mpierr, __FILE__, __LINE__);        
+        return check_mpi2(ios, NULL, mpierr, __FILE__, __LINE__);
     LOG((3, "max_maplen = %d", max_maplen));
 
     /* 2D array that, on task 0, will hold all the map information for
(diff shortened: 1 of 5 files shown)


src/clib/pioc_support.c
```diff
@@ -965,10 +969,100 @@ int PIOc_write_nc_decomp(const char *filename, int iosysid, int ioid, MPI_Comm c
...
+    if ((ret = pioc_read_nc_decomp_int(iosysid, filename, &ndims, &global_dimlen, &num_tasks_decomp,
+                                       &task_maplen, &max_maplen, &full_map, title, history,
+                                       source_in, version_in, fortran_order)))
+        return ret;
+    LOG((2, "ndims = %d num_tasks_decomp = %d max_maplen = %d", ndims, num_tasks_decomp,
+         max_maplen));
+
+    /* If the size does not match the number of tasks in the decomp,
+     * that's an error. */
+    if (size != num_tasks_decomp)
```
jedwards4b: We may want to allow that size is >= to num_tasks_decomp

src/clib/pioc_support.c
```diff
@@ -965,10 +969,100 @@ int PIOc_write_nc_decomp(const char *filename, int iosysid, int ioid, MPI_Comm c
...
+    if (!ret)
+    {
+        PIO_Offset compmap[task_maplen[my_rank]];
+
+        /* Copy array into PIO_Offset array. */
+        for (int e = 0; e < task_maplen[my_rank]; e++)
+            compmap[e] = full_map[my_rank * max_maplen + e];
+
+        /* Initialize the decomposition. */
+        ret = PIOc_InitDecomp(iosysid, basetype, ndims, global_dimlen, task_maplen[my_rank],
```
jedwards4b: It's not apparent to me that we would want to do this here rather than let the user do it?

src/clib/pioc_support.c
edhartnett: What should happen in such a case?

For example if I have a decomp file for 4 tasks, and I read it on 6 tasks, do you want me to just distribute the data on the first 4 tasks and ignore the extra 2?

src/clib/pioc_support.c
edhartnett: One reason is that it is not the custom to allocate arrays and expect the user to free them, and I have to do that here to read the mapping.

My idea is that reading and writing a decomp file should be symetrical. The user should specify an iosystemid, and ioid to write, and specify an iosystemid and get an ioid when reading. Does that make sense?

PR #680, opened 2017-02-19
return value checking, changing parameters to pointers to struct

### LINKED ISSUE #675: incorrect netCDF error handling in some functions in pio_darray_int.c
The netCDF function is called in a loop, but ierr is only checked outside the loop.

pio_read_darray_nc_serial()
pio_read_darray_nc()
pio_write_darray_multi_nc_serial()

### LINKED ISSUE #676: Change flush_buffer() in pio_darray_int.c to return error code
There is a call to PIOc_write_darray_multi() which may return an error code, however this is currently ignored.

### LINKED ISSUE #677: Probably want to pass pointer to ios to some functions in pio_rearrange.c
This function:

```
int rearrange_comp2io(iosystem_desc_t ios, io_desc_t *iodesc, void *sbuf,
                      void *rbuf, int nvars)

```

Takes a copy of the iosystem_desc_t, all it needs is a pointer to it.

This also is the case for:
define_iodesc_datatypes()
subset_rearrange_create()
compute_buffer_init()
free_cn_buffer_pool()
compute_maxaggregate_bytes()
piomemerror()
rearrange_comp2io()
performance_tune_rearranger()
subset_rearrange_create()

### LINKED ISSUE #678: Check return code from define_iodesc_datatypes() in functions in pio_rearranger.c
rearrange_comp2io() 
rearrange_io2comp() 

both call define_iodesc_datatypes() but ignore return codes.

### LINKED ISSUE #679: Check return code from rearrange_comp2io() in PIOc_write_darray_multi()

edhartnett:
In this PR I add return code checking in several places. Remember:

`void functions + programmer laziness = ignored return codes
`
So just return int from all new functions. ;-)

Also changed some functions that were passing whole struct instead of pointer to struct. Using the struct causes the entire contents of the struct to be copied, instead of just a pointer to it. So we need to follow the C convention of always passing pointers to structs, not the structs themselves.

Fixes #679. 
Fixes #678. 
Fixes #676. 
Fixes #677.
Fixes #675.

### DIFF
--- a/doc/source/mach_walkthrough.txt
+++ b/doc/source/mach_walkthrough.txt
@@ -426,7 +426,7 @@ build from the command line, including tests:
 
 <pre>
 cd build
-rm -rf * && CFLAGS=-g FFLAGS=-g CC=mpicc FC=mpifort cmake -DNetCDF_C_PATH=/usr/local/netcdf-4.4.1 -DNetCDF_Fortran_PATH=/usr/local/netcdf-fortran-4.4.4 -DPnetCDF_PATH=/usr/local/pnetcdf-1.7.0 -DPIO_ENABLE_LOGGING=On -DPIO_HDF5_LOGGING=On .. && make VERBOSE=1 all tests check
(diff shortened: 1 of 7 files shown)


jayeshkrishna:
Was this PR reviewed?

edhartnett:
Sorry, I actually thought Jim had approved it, but that was an old email about a different PR, as it turns out.

So please take some time to review now. BTW don't just review changes - if you see anything in the code that looks funny, that's a good thing to bring up, even it if is not part of my changes. This code has not been reviewed much in the past, AFAIK.

rljacob:
Please don't rely on email since there's so many.  Check the "Reviewers" on the top right and see if they did it.

Issue #672, opened 2017-02-17
Change use of ncid variable in darray functions


edhartnett:
Throughout the darray functions we see:

 `ncid = file->fh;`

This made sense before we changed to using an internal vs. external ncid. Now this line makes it confusing as to which ncid is being used. The code should just use file->fh (the internal ncid), or file->pio_ncid (the external ncid) as necessary.

PR #683, opened 2017-02-20
more test

### LINKED ISSUE #681: 21 Fortran tests failed memory leak check
This issue is caused by PR #680, which is reproducible in both develop and master branches.

[Root cause of the leaks]
In flush_buffer(), pio_darray_int.c, the code is updated by the PR such that if PIOc_write_darray_multi() returns an error, flush_buffer() will return that error, without executing the following clean up code.
```
/* If there are any variables in this buffer... */
if (wmb->validvars > 0)
{
    /* Write any data in the buffer. */
    if ((ret = PIOc_write_darray_multi(ncid, wmb->vid,  wmb->ioid, wmb->validvars,
                                       wmb->arraylen, wmb->data, wmb->frame,
                                       wmb->fillvalue, flushtodisk)))
        return pio_err(NULL, file, ret, __FILE__, __LINE__);

    wmb->validvars = 0;

    /* Release the list of variable IDs. */
    brel(wmb->vid);
    wmb->vid = NULL;
    ...
}
```

[Code that triggers the leaks]
On the other hand, in PIOc_write_darray_multi(), pio_darray.c, the code is updated by the PR to check returned value from flush_output_buffer(). For the 21 failed Fortran tests, flush_output_buffer() does return an error code (e.g. PIO_EBADID), which is then returned by PIOc_write_darray_multi().
```
/* Flush data to disk. */
if (file->iotype == PIO_IOTYPE_PNETCDF)
    if ((ierr = flush_output_buffer(file, flushtodisk, 0)))
```

[How to fix]
We should make sure that the clean up code is executed even if PIOc_write_darray_multi() fails.

edhartnett:
Fixes #681.

 Part of #672.

### DIFF
--- a/tests/cunit/test_darray.c
+++ b/tests/cunit/test_darray.c
@@ -114,7 +114,8 @@ int test_darray(int iosysid, int ioid, int num_flavors, int *flavor, int my_rank
 
     /* Use PIO to create the example file in each of the four
      * available ways. */
-    for (int fmt = 0; fmt < num_flavors; fmt++)
+    /* for (int fmt = 0; fmt < num_flavors; fmt++) */
+    for (int fmt = 1; fmt < 2; fmt++)
     {
         /* Create the filename. */
         sprintf(filename, "data_%s_iotype_%d.nc", TEST_NAME, flavor[fmt]);
@@ -148,6 +149,12 @@ int test_darray(int iosysid, int ioid, int num_flavors, int *flavor, int my_rank
 
         if ((ret = PIOc_setframe(ncid, varid, 0)))
             ERR(ret);
+
+        /* These should not work. */
+        if (PIOc_write_darray(ncid + 42, varid, ioid, arraylen, test_data, &fillvalue) != PIO_EBADID)
+            ERR(ERR_WRONG);
+        if (PIOc_write_darray(ncid, varid, ioid + 42, arraylen, test_data, &fillvalue) != PIO_EBADID)
+            ERR(ERR_WRONG);
         
         if ((ret = PIOc_write_darray(ncid, varid, ioid, arraylen, test_data, &fillvalue)))
             ERR(ret);


tests/cunit/test_darray.c
```diff
@@ -114,7 +114,8 @@ int test_darray(int iosysid, int ioid, int num_flavors, int *flavor, int my_rank
 
     /* Use PIO to create the example file in each of the four
      * available ways. */
-    for (int fmt = 0; fmt < num_flavors; fmt++)
+    /* for (int fmt = 0; fmt < num_flavors; fmt++) */
+    for (int fmt = 1; fmt < 2; fmt++)
```
jayeshkrishna: Can you change the number "2" to a constant/variable ?

tests/cunit/test_darray.c
```diff
@@ -148,6 +149,12 @@ int test_darray(int iosysid, int ioid, int num_flavors, int *flavor, int my_rank
 
         if ((ret = PIOc_setframe(ncid, varid, 0)))
             ERR(ret);
+
+        /* These should not work. */
+        if (PIOc_write_darray(ncid + 42, varid, ioid, arraylen, test_data, &fillvalue) != PIO_EBADID)
```
jayeshkrishna: Can you change "42" to a constant/variable that shows what you are doing here?

tests/cunit/test_darray.c
```diff
@@ -148,6 +149,12 @@ int test_darray(int iosysid, int ioid, int num_flavors, int *flavor, int my_rank
...
+            ERR(ERR_WRONG);
+        if (PIOc_write_darray(ncid, varid, ioid + 42, arraylen, test_data, &fillvalue) != PIO_EBADID)
```
jayeshkrishna: Same comment here. "42" needs to be changed to some const/variable

jayeshkrishna:
Why was this PR closed?

edhartnett:
I picked the wrong branch - meant to pick the other one, which is now open. Github doesn't let me delete it, so I closed it.

jayeshkrishna:
ok, thanks. I would be good to add comments when closing PRs without a merge.

PR #684, opened 2017-02-20
took out ncid, fixed memory leak

### LINKED ISSUE #681: 21 Fortran tests failed memory leak check

edhartnett:
Part of #672.
Fixes #681

### DIFF
--- a/src/clib/pio_darray_int.c
+++ b/src/clib/pio_darray_int.c
@@ -478,13 +478,12 @@ int pio_write_darray_multi_nc(file_desc_t *file, int nvars, const int *vid, int 
                               int num_aiotasks, void *iobuf, const int *frame)
 {
     iosystem_desc_t *ios;  /* Pointer to io system information. */
-    var_desc_t *vdesc;
+    var_desc_t *vdesc;     /* Pointer to info about the var. */
     int i;
-    int dsize;
-    int fndims;
-    PIO_Offset tdsize;
+    int fndims;            /* Number of dims for this var in file. */
+    int dsize;             /* Number of elements in data (pnetcdf only). */
+    PIO_Offset tdsize;     /* Number of elements in data (pnetcdf only). */
     int tsize = 0;
-    int ncid;
     int mpierr = MPI_SUCCESS, mpierr2;  /* Return code from MPI function codes. */
     int ierr = PIO_NOERR;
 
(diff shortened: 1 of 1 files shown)


jayeshkrishna:
Did taking out the ncid fix the memory leak?

edhartnett:
Nope, but taking out the ncid was something I had just completed, so I just fixed the memory leak and put my branch up for review.

dqwu:
Next time it would be better to use two PRs, one for taking out the ncid, and one for fixing memory leaks. Try to let each PR make minimal code changes.

jayeshkrishna:
I cannot figure it out from the commit messages (it would be good to have separate commits and detailed commit messages so that it is clear how/what bugs got fixed).
I defer the review to @jedwards4b

edhartnett:
You really can't figure out what is going on in this code? Or you are just enjoying busting on me? 

I will close this PR and break this into two PRs, and you take a look and try again.

jayeshkrishna:
From a quick look I could not figure it out. But if @jedwards is ok with it I am fine too (you don't have to recreate it if @jedwards is ok with it).

edhartnett:
@jayeshkrishna, @dqwu,  I welcome your comments on my code.

What I would like to see is the same interest in following process rules, we all also have for reviewing code and making comments. 

We all want this library to contain the highest quality code, so I invite you all to study the code, get to know it, and point out where it can be improved by removing unneeded complexity, adding documentation, changing variable names to be more clear, etc. And, of course, by finding bugs. ;-)

Thanks,
Ed

PR #699, opened 2017-02-21
use PNET_READ_AND_BCAST always

### LINKED ISSUE #380: What is pre-processor macro PNET_READ_AND_BCAST
In pio_getput_int.c I see:

> #ifdef PNET_READ_AND_BCAST

What is that all about?

### LINKED ISSUE #537: Making PNET_READ_AND_BCAST default in pio_getput_int.c?
In PIOc_get_vars_tc(), we have two code sections for PIO_IOTYPE_PNETCDF. The default is not using PNET_READ_AND_BCAST, unless that macro is explicitly defined by users when building PIO. 
```
#ifdef PNET_READ_AND_BCAST
    LOG((1, "PNET_READ_AND_BCAST"));
    ...
    if (ios->iomaster == MPI_ROOT)
    ...
#else
    LOG((1, "not PNET_READ_AND_BCAST"));
    ...
#endif /* PNET_READ_AND_BCAST */
...
if ((mpierr = MPI_Bcast(buf, num_elem * typelen, MPI_BYTE, ios->ioroot, ios->my_comm)))
```

It seems that using PNET_READ_AND_BCAST is better, that is: only the IO master does the IO.

With current PIO model, all IO tasks in above code will have exactly the same start and count. If we are not using PNET_READ_AND_BCAST, we are actually not utilizing the features of PNETCDF here as the IO tasks will get exactly the same data in the buffer (start and count used are the same). Also, the MPI broadcast will send the same data again from IO master to other IO tasks. So for some IO tasks, they get data from netcdf file first, then get the same data from the IO master.

Using PNET_READ_AND_BCAST does not have this duplicated data issue (less overhead): only the IO master reads the data, and sends it to other IO tasks.

Unless the model is updated such that each IO tasks can have different start and count here (true parallel read), we can change to use PNET_READ_AND_BCAST by default.

edhartnett:
Fixes #537.
Fixes #380.

### DIFF
--- a/src/clib/pio_getput_int.c
+++ b/src/clib/pio_getput_int.c
@@ -646,8 +646,6 @@ int PIOc_get_vars_tc(int ncid, int varid, const PIO_Offset *start, const PIO_Off
 #ifdef _PNETCDF
         if (file->iotype == PIO_IOTYPE_PNETCDF)
         {
-#ifdef PNET_READ_AND_BCAST
-            LOG((1, "PNET_READ_AND_BCAST"));
             ncmpi_begin_indep_data(file->fh);
 
             /* Only the IO master does the IO, so we are not really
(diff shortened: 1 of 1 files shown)


dqwu:
Should we still keep the code block for "not PNET_READ_AND_BCAST" case (can still be enabled by defining a macro)? If it will no longer be used, then it is OK to completely remove it.

edhartnett:
No I don't think we should keep the other code block. The one we are keeping has better performance and passes all tests.

Keeping code is not free. It must be documented, maintained, and tested. Best to eliminate all code we do not need.

Anyway it is always present in the repo if someone finds a reason for it in the future.

PR #718, opened 2017-02-23
Changes to read/write netCDF decomp files

### LINKED ISSUE #436: Add print_trace() to netCDF decomp file as text attribute
It is untested currently. Perhaps this should become part of the logging code, and only be included if logging is enabled...

### LINKED ISSUE #693: Add cmode parameter to PIOc_write_nc_decomp()
Currently the code uses classic netCDF for the new decomp files.

The only problem is that some of our numbers in the decompositon are PIO_Offset (64-bit int), and that type is not available in classic. So 32-bit int is used. This will make decomposition files useless to those using very large dimensions.

Well, this is for supercomputers, so I guess we have to accommodate some very large dimensions. I will change the code to allow the user to specify an iotype for the decomp file. If one of the netCDF-4 iotypes is used, I will change to 64-bit ints.

### LINKED ISSUE #717: Make iosysid first param in PIOc_read/write_nc_decomp()
Usually iosysid is the first parameter...

edhartnett:
Added cmode param to allow netCDF-4 decomp files. (Fixes #693)
Changed parameter order to match other functions. (Fixes #717)
Add backtrace as an attribute. (Fixes #436)
More tests.

### DIFF
--- a/src/clib/pioc_support.c
+++ b/src/clib/pioc_support.c
@@ -259,10 +259,9 @@ void pio_log(int severity, const char *fmt, ...)
  * if using the pio performance tool leads to tuning that could be
  * applied in the model you know more or less where to do it.
  *
- * It's also useful if you have a model bug and used the env variable
- * to enable printing decompositions - then you have 20 or so of them
- * and you need to identify the one that was problematic.  So it's
- * used as an add to the developer and not used at all by any
+ * It's also useful if you have a model bug - then you have 20 or so
+ * decomp files and you need to identify the one that was problematic.
+ * So it's used as an add to the developer and not used at all by any
  * automated process or tools.
  *
  * @param fp file pointer to send output to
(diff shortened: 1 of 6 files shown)


src/clib/pioc_support.c
```diff
@@ -259,10 +259,9 @@ void pio_log(int severity, const char *fmt, ...)
  * if using the pio performance tool leads to tuning that could be
  * applied in the model you know more or less where to do it.
  *
- * It's also useful if you have a model bug and used the env variable
- * to enable printing decompositions - then you have 20 or so of them
- * and you need to identify the one that was problematic.  So it's
- * used as an add to the developer and not used at all by any
+ * It's also useful if you have a model bug - then you have 20 or so
+ * decomp files and you need to identify the one that was problematic.
+ * So it's used as an add to the developer and not used at all by any
```
jedwards4b: add should be aid

PR #720, opened 2017-02-24
fix issue with non iotasks calling flush

### LINKED ISSUE #719: recent commit breaks cesm
I have been testing in cesm, most recently trying to track down a bug in netcdf4 on the ncar cheyenne system.   I recently updated to the head of master and got a new error: 

0:ERROR: 0 NetCDF: Not a valid ID err_num = -33 fname = /glade/scratch/jedwards/forbrian/cesm2_0_alpha06/cime/externals/pio2/src/clib/pio_darray.c line = 261

I've traced this back to commit f9bee9e Merge pull request #680 from NCAR/ejh_ret_codes

I do not get this error in commit 6ad90b0 Merge pull request #651 from NCAR/jayeshkrishna/more_1d_tests

jedwards4b:
Non-iotasks were calling ncmpi functions and thus returning an invalid file id error.
I have merged to develop for testing
Fixes #719

### DIFF
--- a/src/clib/pio_darray_int.c
+++ b/src/clib/pio_darray_int.c
@@ -1521,7 +1521,7 @@ int flush_output_buffer(file_desc_t *file, bool force, PIO_Offset addsize)
 
     /* Find out the buffer usage. */
     if ((ierr = ncmpi_inq_buffer_usage(file->fh, &usage)))
-	return pio_err(ios, file, PIO_EBADID, __FILE__, __LINE__);
+	return pio_err(NULL, file, PIO_EBADID, __FILE__, __LINE__);
 
     /* If we are not forcing a flush, spread the usage to all IO
      * tasks. */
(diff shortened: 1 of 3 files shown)


src/clib/pio_darray_int.c
```diff
@@ -1521,8 +1521,7 @@ int flush_output_buffer(file_desc_t *file, bool force, PIO_Offset addsize)
 
     /* Find out the buffer usage. If I check the turn code, some tests fail. ??? */
     if ((ierr = ncmpi_inq_buffer_usage(file->fh, &usage)))
-        return ierr;
-    /*return pio_err(ios, file, PIO_EBADID, __FILE__, __LINE__);*/
+	return ierr;
```
edhartnett: Can this return code be handled with pio_err?

jedwards4b: Was this pio_err commented out because non-io tasks were getting here?  Perhaps it can be put back in.

edhartnett: That makes a lot of sense!

It was commented out, as indicated by the comment, because when I put it in, some tests failed. I had not yet figured out why, but it could be the same bug you found.

Please check it, because, if true, it means that we already have a test that detects that problem. ;-)

jedwards4b:
I replaced with pio_err, ran tests on my machine to confirm it works and pushed to develop.

edhartnett:
OK if I merge this? It has passed cdash testing...

PR #721, opened 2017-02-24
Allow user buffers larger than local ndof

### LINKED ISSUE #577: PIO fails when user buffer is larger than local ndof size
Currently PIO fails in pio_write_darray() when the user buffer size is
not equal to the expected local ndof size. The user should be able to
pass in a buffer that is larger than the local ndof size (e.g. when the
user is reusing a buffer) as long as it is valid. We should check
if the size of the passed in buffer is at least as large as the expected
(local ndof) size

jayeshkrishna:
Allowing user buffers larger than the size expected (as specified in the
IO decomposition) by the current process.
* Fixes a C test that checked for larger user buffers to test for smaller
  user buffers
* Adds a new Fortran test that specifies larger user buffers
* Minor doc change

Fixes #577

### DIFF
--- a/src/clib/pio_darray.c
+++ b/src/clib/pio_darray.c
@@ -275,10 +275,13 @@ int PIOc_write_darray_multi(int ncid, const int *vid, int ioid, int nvars, PIO_O
  * to.
  * @param ioid the I/O description ID as passed back by
  * PIOc_InitDecomp().
- * @param arraylen the length of the array to be written. This is the
- * length of the local component of the distrubited array. That is,
- * the length of the portion of the data that is on this task. This
- * must be equal to the size of the local array in the decomposition.
+ * @param arraylen the length of the array to be written.
+ * This should be at least the length of the local component of the 
+ * distrubited array. That is, the length of the portion of the data
+ * that is on this task. This must be at least the size of the local
+ * array in the decomposition and the initial elements in the array
+ * should contain the local component of the data, as specified in
+ * the decomposition (ioid).
  * @param array pointer to the data to be written. This is a
  * pointer to the distributed portion of the array that is on this
  * task.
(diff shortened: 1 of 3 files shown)


src/clib/pio_darray.c
```diff
@@ -275,10 +275,13 @@ int PIOc_write_darray_multi(int ncid, const int *vid, int ioid, int nvars, PIO_O
...
  * PIOc_InitDecomp().
- * @param arraylen the length of the array to be written. This is the
- * length of the local component of the distrubited array. That is,
- * the length of the portion of the data that is on this task. This
- * must be equal to the size of the local array in the decomposition.
+ * @param arraylen the length of the array to be written.
+ * This should be at least the length of the local component of the 
+ * distrubited array. That is, the length of the portion of the data
+ * that is on this task. This must be at least the size of the local
+ * array in the decomposition and the initial elements in the array
```
edhartnett: I don't follow the comment. What do you mean "the initial elements in the array should contain the local component of the data".

Should all elements, not just the initial ones, contain the local component of the data?

src/clib/pio_darray.c
```diff
@@ -275,10 +275,13 @@ int PIOc_write_darray_multi(int ncid, const int *vid, int ioid, int nvars, PIO_O
...
+ * should contain the local component of the data, as specified in
+ * the decomposition (ioid).
```
edhartnett: If the user supplies an array that is longer than needed, what happens to the aditional elements? Are they ignored?

src/clib/pio_darray.c
jayeshkrishna: Consider the case were arraylen > number of elements required (as per iodesc) by the current process. In that case we only should be worried about the initial elements (=local ndof) in the array.

src/clib/pio_darray.c
jayeshkrishna: Yes, we should just ignore them

jayeshkrishna:
I will create a separate issue that cleans up the code in pioc_write_darray() wrt the assumptions made about arraylen (when caching data etc)

PR #723, opened 2017-02-24
add darray testing for different types

### LINKED ISSUE #665: Test (non-async) darrays of all types
Needs to be done before #89, #388, #473 and maybe some others. ;-)

edhartnett:
Looking at the code, it seems that the only types that are supported with darray are PIO_INT, PIO_FLOAT, and PIO_DOUBLE.

In this PR I change test_darray.c so that it tests all three types. Only test code changes in this PR.

Future work will include trying out other types. ;-)

Fixes #665.
Part of #205.


edhartnett:
The rest of the netCDF atomic types are coming soon.

If we do it right, we can get netCDF-4 user-defined types working as well. That would be nice!

PR #734, opened 2017-02-26
Fix pointless check in darray functions, added comments

### LINKED ISSUE #715: Useless check in PIOc_write_darray()
I see this useless check:

```
    /* Get var description. */
    if (!(vdesc = file->varlist + vid))
        return pio_err(ios, file, PIO_EBADID, __FILE__, __LINE__);
    LOG((2, "vdesc record %d ndims %d nreqs %d", vdesc->record, vdesc->ndims, vdesc->nreqs));

```

What would be useful would be to check that vid is valid, if it can be done in a very efficient way.

edhartnett:
Fixes #715.


src/clib/pio_darray_int.c
```diff
@@ -490,10 +488,11 @@ int pio_write_darray_multi_nc(file_desc_t *file, int nvars, const int *vid, int
     GPTLstart("PIO:write_darray_multi_nc");
 #endif
 
-    /* Get file and variable info. */
+    /* Get pointer to iosystem. */
     ios = file->iosystem;
-    if (!(vdesc = file->varlist + vid[0]))
-        return pio_err(ios, file, PIO_EBADID, __FILE__, __LINE__);
+
+    /* Point to var description scruct for first var. */
```
jedwards4b: struct

src/clib/pio_darray_int.c
```diff
@@ -1434,12 +1466,15 @@ int pio_read_darray_nc_serial(file_desc_t *file, io_desc_t *iodesc, int vid,
...
 
+                /* Now get each region of data. */
                 loffset = 0;
                 for (int regioncnt = 0; regioncnt < maxregions; regioncnt++)
                 {
+                    /* Get pointer where data should go. */
                     bufptr = (void *)((char *)iobuf + tsize * loffset);
                     regionsize = 1;
 
+                    /* ??? */
```
jedwards4b: The decomposition may not use all of the active io tasks rtask here is the io task rank and ios->num_iotasks is the number of iotasks actually used in this decomposition.

edhartnett: OK, thanks, I will add this comment in my next PR...

PR #738, opened 2017-02-27
Fix netcdf read error across rearrangers

### LINKED ISSUE #737: NetCDF read fails with subset rearranger after a box rearranger is used
In the same program if the subset rearranger is used after the box rearranger to read
data, the read is not completed and we get invalid values in the read buffer
 (the read buffer is not updated).
This error is due to mismatch in the recv and sends when reading data in serial.

jayeshkrishna:
Fixing NetCDF read failures across rearrangers. Also adding some
tests for rearrangers.

Fixes #737


jayeshkrishna:
Did the merge require a "manual merge" ?

jayeshkrishna:
If the merge required a manual merge I would prefer that it be done by the developer working on the issue (who has been "assigned" the PR).

PR #762, opened 2017-03-01
Fix issues in flow control code and associated tests

### LINKED ISSUE #637: pio_swapm() test coverage shows that it is not being tested
We have a function pio_swapm() in pio_spmd.c.

It is like MPI_Alltoallw() with flow control.

But in our tests, we never use the flow control. I'm not sure if we do in the code either. (I believe this is an intended optimization that has not yet be achieved.)

Is this similar to the flow control that Jayesh just added? Or is that different?

### LINKED ISSUE #645: is documentation of ceil2() function in pio_spmd.c correct?

jayeshkrishna:
This PR includes,
* Fixes to the fortran pio rearranger tests to use all the rearranger options
* More Fortran tests for testing rearranger options
* Fixes to the pio flow control code
* Minor doc fixes and logging

Fixes #637 
Fixes #645

### DIFF
--- a/src/clib/pio_spmd.c
+++ b/src/clib/pio_spmd.c
@@ -321,7 +321,7 @@ int pio_swapm(void *sendbuf, int *sendcounts, int *sdispls, MPI_Datatype *sendty
 
         /* We did comms in sets of size max_reqs, if istep > maxreqh
          * then there is a remainder that must be handled. */
-        if (istep > maxreqh)
+        if (istep > maxreqh - 1)
         {
             p = istep - maxreqh;
             if (rcvids[p] != MPI_REQUEST_NULL)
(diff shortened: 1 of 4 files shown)


src/clib/pio_spmd.c
```diff
@@ -291,7 +302,7 @@ int pio_swapm(void *sendbuf, int *sendcounts, int *sdispls, MPI_Datatype *sendty
              * used to choose between mpi_irsends and mpi_isends - the default
              * is still mpi_irsend
              */
-            if (isend)
+            if (handshake && isend)
```
jedwards4b: This change doesn't make any sense to me.  The code for isend here is the same whether or not handshake is true, except that you removed the ifdef USE_MPI_ISEND_FOR_FC

jayeshkrishna: I don't think we can guarantee that a recv is posted without an application handshake (since we post recvs in blocks).

PR #804, opened 2017-03-03
new, less confusing PIOc_init_async() API

### LINKED ISSUE #402: Clean up API and documentation of PIOc_Init_Async()
In pioc.c, PIOc_Init_Async() has a description for its parameters:
```
 * @param num_procs_per_comp an array of int, of length
 * component_count, with the number of processors in each computation
 * component.

 * @param proc_list an array of arrays containing the processor
 * numbers for each computation component.
```
From the function body and unit tests that calls this function, num_procs_per_comp is of length component_count + 1 (including an IO component before computational components).
The example below is from test_async_3proc.c
```
 /* Num procs for IO and computation. */
 int num_procs[NUM_COMBOS][COMPONENT_COUNT + 1] = {{2, 1}, {1, 2}};
```

For proc_list, it should also include an IO component. When the code creates one if the user did not provide it, the IO component is counted (component_count plus 1)
```
/* Allocate space for array of arrays. */
if (!(my_proc_list = malloc((component_count + 1) * sizeof(int *))))
    return pio_err(NULL, NULL, PIO_ENOMEM, __FILE__, __LINE__);
```

edhartnett:
The API to PIOc_Init_Async() is unclear and poorly documented.

This PR introduces new function PIOc_init_async(). Yes the name is confusing bu PIOc_Init_Async() will be taken out in a future PR pretty shortly.

This is the same code, but now the API makes a more clear separation between IO and computation components.

I'm just putting this up to do the change in smaller batches, and to give people time to comment on the changed API before I change it everywhere in the library and tests. With this PR one of the async tests has been converted to the new PIOc_init_async().

Fixes #402.

### DIFF
--- a/src/clib/pioc.c
+++ b/src/clib/pioc.c
@@ -1687,26 +1687,6 @@ int PIOc_init_async(MPI_Comm world, int num_io_procs, int *io_proc_list,
         in_cmp = (pidx == num_procs_per_comp[cmp]) ? 0 : 1;
         LOG((3, "pidx = %d num_procs_per_comp[%d] = %d in_cmp = %d",
              pidx, cmp, num_procs_per_comp[cmp], in_cmp));
-
-        /* Create an intracomm for this component. Only processes in
-         * the component need to participate in the intracomm create
-         * call. */
-        /* Create the intracomm from the group. */
-        LOG((3, "creating intracomm cmp = %d from group[%d] = %d", cmp, cmp, group[cmp]));
-
-        /* We handle the IO comm differently (cmp == 0). */
-        if (!cmp)
-        {
-            /* LOG((3, "about to create io comm")); */
-            /* if ((ret = MPI_Comm_create_group(world, group[cmp], cmp, &io_comm))) */
-            /*     return check_mpi(NULL, ret, __FILE__, __LINE__); */
-            /* LOG((3, "about to get io rank"));                 */
-            /* if ((ret = MPI_Comm_rank(io_comm, &io_rank))) */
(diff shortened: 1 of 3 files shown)


src/clib/pioc.c
```diff
@@ -1451,6 +1451,430 @@ int PIOc_Init_Async(MPI_Comm world, int num_io_procs, int *io_proc_list,
...
+        my_iosys->my_comm = MPI_COMM_NULL;
+        my_iosys->async_interface = 1;
+        my_iosys->error_handler = default_error_handler;
+        my_iosys->num_comptasks = num_procs_per_comp[cmp];
+        my_iosys->num_iotasks = num_io_procs;
+        my_iosys->compgroup = MPI_GROUP_NULL;
+        my_iosys->iogroup = MPI_GROUP_NULL;
+        
+        /* The rank of the computation leader in the union comm. */
+        my_iosys->comproot = num_io_procs;
```
jedwards4b: I think that there is something wrong here - are you assuming that this routine will define the comms for each component?  and that the io tasks will consist of tasks 0, num_io_tasks-1 of comm world?   I think that these are generally bad assumptions - the application task should define and pass in it's comp_comms along with world 
and the io_comm should be what's left of world when all the comp_comms are removed.

edhartnett: OK, noted. Let me ponder that. However I will still move forward with this PR because it is an improvement on the current code.

jedwards4b: Looking back at the fortran implementation - the application defines the compute comm for each component and the io_comm and passes them all in along with world and the pio init function creates the union comms.

PR #846, opened 2017-03-08
Adding some fortran tests that puts/gets slices of arrays

jayeshkrishna:
Adding two tests that puts/gets slices of arrays

* A test that puts/gets a 1d array slice of a 2d var
* A test that puts time slices of 2d/3d/4d vars and reads it


jayeshkrishna:
Some similar tests (for the 4d arrays) for pio_darray() functions are already there in tests/general/pio_decomp_frame_tests.F90.in

jedwards4b:
Great thanks

PR #850, opened 2017-03-08
need this to use check_function_exists

jedwards4b:
I don't understand cmake well enough to puzzle this out, but sometimes we get an error that
check_function_exists doesn't exist.   After adding this include I have not been able to reproduce that error.     I will merge to develop

### DIFF
--- a/cmake/LibCheck.cmake
+++ b/cmake/LibCheck.cmake
@@ -1,5 +1,5 @@
 include (CMakeParseArguments)
-
+include (CheckFunctionExists)
 #==============================================================================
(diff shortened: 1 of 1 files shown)


jayeshkrishna:
I thought @edhartnett had fixed this before

edhartnett:
Similar, but different. I added a similar line to src/clib/CMakeLists.txt:
`include(CheckTypeSize)`

This was to ensure that the size of long long and size_t are the same, otherwise lots of our casts would be very very wrong. ;-)

PR #879, opened 2017-03-10
fixed if/ifdef issue, stopped some async tests that fail for bget

### LINKED ISSUE #877: PIO_USE_MALLOC is always applied in bget.c
The option PIO_USE_MALLOC is turned off by default:
`option (PIO_USE_MALLOC       "Use native malloc (instead of bget package)"  OFF)`

Based on this option, in config.h we will have a macro defined:
`#define PIO_USE_MALLOC 0`
or
`#define PIO_USE_MALLOC 1`

However, in bget.c, this macro is incorrectly tested (it is always defined, no matter the value is 0 or 1)
```
#include <config.h>
...
#ifdef PIO_USE_MALLOC
#include <stdlib.h>
#endif
...
#ifdef PIO_USE_MALLOC
    buf = malloc(requested_size);
    return(buf);
#endif
...
```
In pio_darray_int.c, the macro is correctly tested (check whether the value of the macro is 0 or 1, instead of whether it is defined or not)
```
#if !PIO_USE_MALLOC
...
#endif
```
[Some comments]
1) Before PR #122, config.h is not included in bget.c, so the macro is not defined at that time in that source file.
2) For now, suppose the option PIO_USE_MALLOC is OFF by default when we run CMake, in bget.c native malloc is always used. However, in pio_darray_int.c, native malloc is not used. This inconsistency might cause some problems.
3) This issue affects test coverage. As we can see in bget.c, bget memory management will never be used
4) This means we have not been testing bget memory management since PR #122.

[How to fix]
The fix is straightforward, just update bget.c to test the value of PIO_USE_MALLOC as in pio_darray_int.c

[Caution]
After the fix is applied, a lot of unit tests will fail. This is expected, as we have not tested them since PR #122. Also, some cime_developer tests will fail if we apply this updated PIO2 to ESMCI CIME. We should continue to fix the failing tests.

edhartnett:
Fixes #877.

Merged to develop for testing.

### DIFF
--- a/src/clib/bget.c
+++ b/src/clib/bget.c
@@ -400,7 +400,7 @@
 #include <config.h>
 #include <pio.h>
 #include <pio_internal.h>
-#ifdef PIO_USE_MALLOC
+#if PIO_USE_MALLOC
 #include <stdlib.h>
 #endif
 
@@ -614,7 +614,7 @@ bufsize requested_size;
     int compactseq = 0;
 #endif
 
-#ifdef PIO_USE_MALLOC
+#if PIO_USE_MALLOC
     //    if(requested_size>maxsize){
     //   maxsize=requested_size;
     //   printf("%s %d %d\n",__FILE__,__LINE__,maxsize);
@@ -847,7 +847,7 @@ bufsize size;
     bufsize osize;                    /* Old size of buffer */
     struct bhead *b;
 
-#ifdef PIO_USE_MALLOC
+#if PIO_USE_MALLOC
     return(realloc(buf, size));
 #endif
     if ((nbuf = bget(size)) == NULL) { /* Acquire new buffer */
@@ -882,7 +882,7 @@ void *buf;
 {
     struct bfhead *b, *bn;
 
-#ifdef PIO_USE_MALLOC
+#if PIO_USE_MALLOC
     //    printf("bget free %d %x\n",__LINE__,buf);
     free(buf);
     return;
--- a/src/clib/pio_darray_int.c
+++ b/src/clib/pio_darray_int.c
@@ -25,6 +25,16 @@ extern void *CN_bpool;
 /* Maximum buffer usage. */
 extern PIO_Offset maxusage;
 
+/* handler for freeing the memory buffer pool */
+void bpool_free(void *p)
+{
+  free(p);
+  if(p == CN_bpool){
+    CN_bpool = NULL;
+  }
+}
+
+
 /**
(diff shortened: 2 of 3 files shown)


dqwu:
Looks good, but there are new memory leaks found in some tests (this is expected as we have not been testing bget package since October 18, 2016).

edhartnett:
I will take a look and fix memory leaks...

dqwu:
@edhartnett 
The memory leaks issue might be tricky. For me, it looks like that for some tests (66 tests), the free on CN_bpool is already called before inside brel(). For some tests (20 tests), the free on  CN_bpool is not called inside brel().

dqwu:
@edhartnett 
Please check any of the following tests. It seems that brel() is never called for these 20 tests that have new memory leaks.
1 - test_spmd
7 - test_iosystem2_simple
8 - test_iosystem2_simple2
9 - test_iosystem2
10 - test_iosystem3_simple
11 - test_iosystem3_simple2
12 - test_iosystem3
25 - init_finialize_1_proc
26 - init_finialize_2_proc
27 - init_finalize_2_proc_with_args
28 - pio_file_simple_tests
29 - pio_file_fail
30 - ncdf_simple_tests
31 - ncdf_get_put_1proc
32 - ncdf_get_put_2proc
33 - ncdf_fail
34 - ncdf_inq
86 - pio_iosystem_tests
87 - pio_iosystem_tests2
88 - pio_iosystem_tests3

For these 66 tests failed earlier, brel() is called, which implicitly calls free(CN_bpool).

edhartnett:
I am not going to merge this PR. I will recreate develop from master this morning as well, so it reflects the code without this change.

I am still working on this bug, but it's a toughie. I am attempting to simplify the situation by accelerating a change that was already underway in the code, as discussed with Jim and approved in a previous meeting: only using bget for our big data needs (i.e. to buffer user data), not using it as memory for every struct.

I will submit some PRs cleaning up various errors of memory handling, and using calloc instead of bget for structs. This will allow me to narrow the use of bget and make it easier to figure out what is going on.

jedwards4b:
Wait I was just about to push a commit that solves the problem.

jedwards4b:
I have pushed the fix to this branch and merged to develop

jayeshkrishna:
One thing missing in the PR is the change to move bpoolrelease() outside the "if(CN_bpool) " branch in free_cn_buffer_pool().
Also need to get rid of passing CN_bpool as arg to bpoolrelease (that does not take any args).

PR #887, opened 2017-03-13
Update CTestEnvironment-anlworkstation.cmake on PIO_USE_MALLOC

dqwu:
This allows CTest scripts for ANL workstations to optionally turn on PIO_USE_MALLOC.

Merged to develop for testing.


jayeshkrishna:
@dqwu : Can you add this option to the fortran tests too?

dqwu:
@jayeshkrishna 
As long as we use -DPIO_USE_MALLOC=ON (default is OFF) when we run CMake command, the generated config.h will have the line "#define PIO_USE_MALLOC 1". This header file is included by bget.c and pio_darray_int.c to use bget package (instead of native malloc). All the Fortran tests will be affected automatically.

jayeshkrishna:
yeah, sorry about that, I thought you had updated files in cunit . please ignore my comments

Issue #888, opened 2017-03-13
memory leak in PIO_USE_MALLOC builds


edhartnett:
There is a memory leak issue that involves this code from PIOc_write_darray_multi():

```
        /* Get a buffer. */
	if (ios->io_rank == 0)
	    vdesc0->fillbuf = bget(iodesc->maxholegridsize * vsize * nvars);
	else if (iodesc->holegridsize > 0)
	    vdesc0->fillbuf = bget(iodesc->holegridsize * vsize * nvars);

```

For pnetcdf cases this memory is released in  flush_output_buffer():

```
        /* Release resources. */
        for (int i = 0; i < PIO_MAX_VARS; i++)
        {
            vdesc = file->varlist + i;
            if (vdesc->iobuf)
            {
                brel(vdesc->iobuf);
                vdesc->iobuf = NULL;
            }
            if (vdesc->fillbuf)
            {
                brel(vdesc->fillbuf);
                vdesc->fillbuf = NULL;
            }
        }

``` 

However the code to free the resources is only called when the buffer is actually flushed. Writes that don't cause a flush leak memory.

This can be seen in test_darray_1d.c.

PR #889, opened 2017-03-13
making tests pass with malloc

### LINKED ISSUE #885: Now some tests fail with USE_MALLOC...
Well I hope and believe are are circling in to the correct solution. ;-)

Our bget based builds are now working OK, but the malloc build is failing. I will take a look...

edhartnett:
This allows tests to pass.

Fixes #885.

This also uncovered a memory leak. See #888.


jedwards4b:
I think that #890 makes this unnecessary.   Please close if you agree.

PR #890, opened 2017-03-13
Fix malloc mem issue

### LINKED ISSUE #885: Now some tests fail with USE_MALLOC...

### LINKED ISSUE #888: memory leak in PIO_USE_MALLOC builds

jedwards4b:
Fixes #885 
Fixes #888
Flush pnetcdf output buffer before attempting to write a new record.  Allocate a token iobuf on all iotasks so that this flush is collective.

### DIFF
--- a/src/clib/pio_darray.c
+++ b/src/clib/pio_darray.c
@@ -152,6 +152,12 @@ int PIOc_write_darray_multi(int ncid, const int *varids, int ioid, int nvars,
                 for (int i = 0; i < iodesc->maxiobuflen; i++)
                     memcpy(&((char *)vdesc0->iobuf)[vsize * (i + nv * iodesc->maxiobuflen)],
                            &((char *)fillvalue)[nv * vsize], vsize);
+    }else if(file->iotype == PIO_IOTYPE_PNETCDF){
+	/* this assures that iobuf is allocated on all iotasks thus assuring that the flush_output_buffer call
+	 above is called collectively (from all iotasks) */
+        if (!(vdesc0->iobuf = bget(1)))
+            return pio_err(ios, file, PIO_ENOMEM, __FILE__, __LINE__);
+        LOG((3, "allocated %lld bytes for variable buffer", (size_t)rlen * (size_t)vsize));
     }
 
     /* Move data from compute to IO tasks. */
--- a/src/clib/pio_darray.c
+++ b/src/clib/pio_darray.c
@@ -116,10 +116,14 @@ int PIOc_write_darray_multi(int ncid, const int *varids, int ioid, int nvars,
      * then move that data one node at a time to the io master node
      * and write (or read). The buffer size on io task 0 must be as
      * large as the largest used to accommodate this serial io
-     * method.  For pnetcdf previous iobufs may not yet be flushed, it's okay to move this pointer*/
+     * method.  */
     vdesc0 = file->varlist + varids[0];
-    if (file->iotype != PIO_IOTYPE_PNETCDF)
-	pioassert(!vdesc0->iobuf, "buffer overwrite",__FILE__, __LINE__);
+
+    /* if the buffer is already in use in pnetcdf we need to flush first */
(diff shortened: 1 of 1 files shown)


src/clib/pio_darray.c
```diff
@@ -202,14 +213,14 @@ int PIOc_write_darray_multi(int ncid, const int *varids, int ioid, int nvars,
...
-        if (vdesc0->fillbuf)
-            piodie("Attempt to overwrite existing buffer",__FILE__,__LINE__);
+	pioassert(!vdesc0->fillbuf, "buffer overwrite",__FILE__, __LINE__);
 
         /* Get a buffer. */
-	if (ios->io_rank == 0)
-	    vdesc0->fillbuf = bget(iodesc->maxholegridsize * vsize * nvars);
-	else if (iodesc->holegridsize > 0)
-	    vdesc0->fillbuf = bget(iodesc->holegridsize * vsize * nvars);
+	if (!vdesc0->fillbuf)
```
jedwards4b: Don't need this line, i'll remove.

jedwards4b:
I've merged to develop

dqwu:
@jedwards4b 
This fix works, but I have found another issue in pio_darray.c:
```
    /* Find out how much free, contiguous space is available. */
    bfreespace(&totfree, &maxfree);

    /* maxfree is the available memory. If that is < 10% greater than
     * the size of the current request needsflush is true. */
    if (needsflush == 0)
        needsflush = (maxfree <= 1.1 * (1 + wmb->validvars) * arraylen * tsize);
```
When PIO_USE_MALLOC is ON (#define PIO_USE_MALLOC 1 in config.h), bfreespace call is meaningless: freelist will not be used, so totfree calculated will be 0 and maxfree calculated will be -1.

This will always set needsflush to 1, performing necessary flush at the beginning. This has caused some tests to fail before this fix.

If PIO_USE_MALLOC is ON, is there a way to calculate free memory space (maybe use some system functions available from C library)?

dqwu:
@jedwards4b @edhartnett 
This PR has caused 21 tests (in develop branch) to fail (only memory leaks) in LeakSanitizer build (Valgrind build not run yet but it should reproduce the same issue), see
http://my.cdash.org/viewTest.php?onlyfailed&buildid=1152878

All the leaks are from pio_darray.c, line 158
```
if (!(vdesc0->iobuf = bget(1)))
    return pio_err(ios, file, PIO_ENOMEM, __FILE__, __LINE__);
```
Please try to fix the leaks in this PR before merged to master.

edhartnett:
I am going to suggest that we merge these changes and continue working on the memory leaks in additional PRs.

Let's remember that these leaks are not new bugs, introduced by this PR, and therefore to be cleaned up in this PR. These are long-standing leaks that we are just starting to detect. (Note that we just started valgrind testing recently, and for the entire time we have been doing it, we have been doing it on the bget build. So there has never been valgrind testing of the malloc build until now.)

So lets merge in these changes and continue working on these bugs.

PR #894, opened 2017-03-14
added test for GDCblocksize

### LINKED ISSUE #893: C test for GCDblocksize() in pioc_sc.c
Need a C test for this.

edhartnett:
Added test and some documentation for this function. Still not clear what is going on - any comments where ??? are would be most appreciated.

Fixes #893.

### DIFF
--- a/src/clib/pioc_sc.c
+++ b/src/clib/pioc_sc.c
@@ -74,11 +74,11 @@ int gcd_array(int nain, int *ain)
 }
 
 /**
- * Return the gcd of nain and any value in ain as int_64.
+ * Return the greatest common devisor of array ain as int_64.
  *
- * @param main
- * @param ain
- * @returns
+ * @param nain number of elements in ain.
+ * @param ain array of length nain.
+ * @returns GCD of elements in ain.
  */
 long long lgcd_array(int nain, long long *ain)
 {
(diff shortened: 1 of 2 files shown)


src/clib/pioc_sc.c
```diff
@@ -177,52 +186,60 @@ PIO_Offset GCDblocksize(int arrlen, const PIO_Offset *arr_in)
...
     }
 
+    /* ??? */
     numblks = numtimes + 1;
     if (numtimes == 0)
         n = numblks;
     else
         n = numtimes;
 
+    /* ??? */
```
jedwards4b: If numblks==1 then the result is arrlen and you can return.

src/clib/pioc_sc.c
```diff
@@ -177,52 +186,60 @@ PIO_Offset GCDblocksize(int arrlen, const PIO_Offset *arr_in)
...
     bsize = (PIO_Offset)arrlen;
     if (numblks > 1)
     {
         PIO_Offset blk_len[numblks];
         PIO_Offset gaps[numtimes];
-        if(numtimes > 0)
+        if (numtimes > 0)
```
jedwards4b: If numblks > 1 then numtimes must be > 0 and this if block isn't needed.

src/clib/pioc_sc.c
```diff
@@ -177,52 +186,60 @@ PIO_Offset GCDblocksize(int arrlen, const PIO_Offset *arr_in)
...
         {
             blk_len[i] = loc_arr[i] - loc_arr[i - 1];
             blklensum += blk_len[i];
         }
         blk_len[numblks - 1] = arrlen - blklensum;
 
+        /* Get the GCD in blk_len array. */
         bsize = lgcd_array(numblks, blk_len);
+
+        /* ??? */
```
jedwards4b: I don't recall why i needed these next two blocks, I remember struggling to get this right in all cases and I'm afraid that the end result is that bsize is almost always 1.

edhartnett:
OK, thanks for the comments! I will add them to the PR...

PR #899, opened 2017-03-14
Memleak fix

jedwards4b:
Tested with and without PIO_USE_MALLOC


jedwards4b:
It looks like someone already pushed this to develop for me?

PR #913, opened 2017-03-16
adding rearranger parameter to init_async function

### LINKED ISSUE #907: Need to add rearranger as part of init_async args
Need to pass in rearranger when doing an init_async, to be the default rearranger.

Part of #89.

### LINKED ISSUE #908: PIOc_init_decomp() can take rearranger as an int, not a pointer
Will not affect old version of this function. 

Part of #89.

edhartnett:
One thing that is preventing some of the decompositon tests from working in async is that async does not handle rearrangers (at all!). So when the code needs a rearranger, it can't find one.

This PR adds the rearranger parameter to the init_async call to get a default rearranger.

Fixes #908.
Fixes #907.
Part of #89.

### DIFF
--- a/src/clib/pio.h
+++ b/src/clib/pio.h
@@ -719,7 +719,7 @@ extern "C" {
 
     /* Init decomposition with 0-based compmap array. */
     int PIOc_init_decomp(int iosysid, int basetype, int ndims, const int *dims, int maplen,
-                         const PIO_Offset *compmap, int *ioidp, const int *rearranger,
+                         const PIO_Offset *compmap, int *ioidp, int rearranger,
                          const PIO_Offset *iostart, const PIO_Offset *iocount);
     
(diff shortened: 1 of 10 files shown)


src/clib/pio.h
```diff
@@ -719,7 +719,7 @@ extern "C" {
 
     /* Init decomposition with 0-based compmap array. */
     int PIOc_init_decomp(int iosysid, int basetype, int ndims, const int *dims, int maplen,
-                         const PIO_Offset *compmap, int *ioidp, const int *rearranger,
+                         const PIO_Offset *compmap, int *ioidp, int rearranger,
```
jedwards4b: Doesn't this make the rearranger a required argument?   If the user sets a default rearranger in PIOc_init_Int* then they shouldn't need to pass it again here.

edhartnett: They pass 0 for the default. It's in the docs.

PR #933, opened 2017-03-21
Fix problem with PIOc_inq_var_fill()

### LINKED ISSUE #930: PIOc_inq_var_fill() not working properly for netCDF classic
Trying to call on all procs, instead of just proc 0.

edhartnett:
Fixes #930.

Also parameter no_fill was misnamed fill_mode.


edhartnett:
In line 2128 the condition file->do_io is added. That's needed so that only IO task 0 attempts to run that code.

The rest is just changing the parameter name to the correct value.

If you have seen some of this code before, it's because I tried to make some changes work this morning, but things got tangled up and some tests started failing. So now I am making the changes one at a time to make sure everything works each step of the way. ;'-)

Issue #857, opened 2017-03-09
C tests for functions in pio_rearrange.c


edhartnett:
The following functions need C tests:
init_rearr_opts()
idx_to_dim_list()
expand_region()
coord_to_lindex()
compute_maxIObuffersize()
define_iodesc_datatypes()
compute_counts()
rearrange_comp2io()
rearrange_io2comp()
determine_fill()
box_rearrange_create()
get_start_and_count_regions()
default_subset_partition()
subset_rearrange_create()

edhartnett:
Tests for most of the listed functions have been added.

Remaining functions to test:

rearrange_comp2io()
rearrange_io2comp()
box_rearrange_create()
default_subset_partition()
subset_rearrange_create()

edhartnett:
With my latest PR, we now have C tests for all these functions.

However in some cases, not much is checked after the test code is run. That is, the function runs and returns 0, but I don't check anything else.

In some cases only one rearranger is tested, but both need to be.

I will take one more pass at improving these tests before closing this ticket.

edhartnett:
Latest coverage report shows good coverage of this code.

PR #936, opened 2017-03-22
added test for default_subset_partition()

### LINKED ISSUE #937: In create_mpi_datatypes() parameter msgcnt must be > 0
Assert it.

edhartnett:
Added test for default_subset_partition().
Also starting testing for rearrange_comp2io().

Part of #857.
Part of #205.
Fixes #937.

### DIFF
--- a/src/clib/pio_internal.h
+++ b/src/clib/pio_internal.h
@@ -152,6 +152,9 @@ extern "C" {
     int compute_counts(iosystem_desc_t *ios, io_desc_t *iodesc, int maplen,
                        const int *dest_ioproc, const PIO_Offset *dest_ioindex,
                        MPI_Comm mycomm);
+
+    /* Compute mapping for subset rearranger. */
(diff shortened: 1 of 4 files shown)


src/clib/pio_internal.h
```diff
@@ -152,6 +152,9 @@ extern "C" {
     int compute_counts(iosystem_desc_t *ios, io_desc_t *iodesc, int maplen,
                        const int *dest_ioproc, const PIO_Offset *dest_ioindex,
                        MPI_Comm mycomm);
+
+    /* Compute mapping for subset rearranger. */
+    int default_subset_partition(iosystem_desc_t *ios, io_desc_t *iodesc);
```
jedwards4b: The as yet unrealized vision here is that the user would be able to supply an alternative subset partitioning function.  Requirements of this function are that there be exactly one io task per compute task group.

edhartnett: Thanks I have added that comment to the code. Keep them coming. Lots of ??? in pio_rearranger.c...

PR #965, opened 2017-04-04
move mpi discovery to top level

jedwards4b:
Have a look and see if this is better than hardcoding 8.   I'll wait for review before merging to develop.


jedwards4b:
Merged to develop

PR #991, opened 2017-04-11
Merge CIME changes back to pio1

jayeshkrishna:
Merge changes back from ESMCI/cime to PIO1. Also fix the 
PIO1 configure (updated in ESMCI/cime) to detect PnetCDF
correctly.

WARNING: This PR needs to be merged after PR #990


pio/CMakeLists.txt
```diff
@@ -31,38 +30,40 @@ endif()
 
 # Netcdf is required
 
+SET (CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../../pio2/cmake" ${CMAKE_MODULE_PATH})
```
jedwards4b: This line seems wrong, CMAKE_MODULE_PATH was already set at line 12

jayeshkrishna: Yes, I think we should open a different issue on this. In this PR I am just copying over what is already in ESMCI/cime. I also don't think setting the CMAKE_MODULE_PATH to ${CMAKE_CURRENT_SOURCE_DIR}/../../pio2/cmake is the right thing to do (shouldn't the makefile set the USER_CMAKE_MODULE_PATH instead).

jayeshkrishna: Once this PR goes in I can create an issue to cleanup CMakeLists.txt in pio1 (and make sure that ESMCI/cime works with this version of pio1)

edhartnett:
Not sure if I should merge this PR or not. Jim, are you OK with Jayesh's plans?

jayeshkrishna:
Please don't merge it until Jim accepts it.

jedwards4b:
Yes I'm okay with this plan.

PR #1000, opened 2017-04-19
need workaround for openmpi

jedwards4b:
In openmpi 2.0.2 the call to mpi_alltoallw in pio_swapm doesn't work correctly - it appears to ignore the offsets in the sending mpi datatype.   This is a workaround that avoids mpi_alltoallw for openmpi,  I will try to create a test case and file a bug with the openmpi developers.

Merged to develop for testing (all tests now passing with openmpi)


jedwards4b:
So this problem gets stranger - I wrote a test program to send to openmpi developers, but before I sent it I was able to test it on another platform using the same openmpi-2.0.2.   But it works there, so maybe we should base this workaround on the particular machine rather than blaming openmpi?  Still researching.

jayeshkrishna:
openmpi might have platform-specific enhancements (network used etc) that might be buggy. I would recommend trying it on your laptop (no fancy networks etc) first.

edhartnett:
Do you want to merge this PR?

jedwards4b:
It seems to be a platform specific problem, 
recompiled openmpi statically on this platform and it no longer has the problem.

jedwards4b:
Okay this PR just updates the module load for the test on hobart now.   Merged to development.

jedwards4b:
I thought that we had everything working, but the latest tests with the static build are behaving just like the shared lib build did...

edhartnett:
Do you want me to merge this PR before the release?

jedwards4b:
No this just affects the cdash testing, no need to merge it.

PR #1010, opened 2017-04-27
Switch default from MPI_Irsends to MPI_Isends

### LINKED ISSUE #946: Use MPI_[I]send instead of MPI_[I]Rsend by default
At least on some MPI implementations (cray-mpich on titan) mpi_[i]rsend
is buggy and results in hangs during I/O.
We can change the default from rsends to isends and leave a compile-time
switch back to using rsends if needed.
This change needs to be done for both PIO1 and PIO2.

jayeshkrishna:
On some systems MPI_Irsends are buggy and can cause a hang (we
found this on titan with pio1). So switching the default from
using MPI_Irsends to MPI_Isends.

The user can use USE_MPI_IRSEND_FOR_FC (compile time flag)
to choose MPI_Irsend instead of the default.

Fixes #946


jayeshkrishna:
See related discussions/concerns in #1011

jayeshkrishna:
Will add custom flags in ACME makefile/config_machines to switch to isends.

PR #1011, opened 2017-04-27
Change PIO1 default from MPI_[I]Rsends to MPI_[I]sends

### LINKED ISSUE #946: Use MPI_[I]send instead of MPI_[I]Rsend by default

jayeshkrishna:
On some systems MPI_Irsends are buggy and can cause a hang (we
found this on titan with pio1). So switching the default from
using MPI_Irsends to MPI_Isends.

The user can use "_USE_MPI_RSEND" (compile time flag) to choose
MPI_[I]Rsends instead of the default

Fixes #946

### DIFF
--- a/pio/pio_spmd_utils.F90.in
+++ b/pio/pio_spmd_utils.F90.in
@@ -19,7 +19,7 @@
 ! Code added as a work around for poor rsend performance on cray systems with
 ! Gemini interconnect
 !
-#ifdef _NO_MPI_RSEND
+#ifndef _USE_MPI_RSEND
 #define MPI_RSEND MPI_SEND
 #define mpi_rsend mpi_send
 #define MPI_IRSEND MPI_ISEND


jayeshkrishna:
MPI_[I]Rsend is a less used (hence likely less tested) API and debugging these kinds of hangs has been difficult. It would be better to use a more reliable API as the default (and leave more optimized routines as a choice).
Do we gain a lot in performance by using MPI_[I]Rsends instead of plain MPI_Isends?

jedwards4b:
The reason we use rsend is to indicate to the MPI library that the code is structured to allow for an optimization here if the mpi library were able to exploit one.  Realized performance, although appreciated, is not really the point.

jayeshkrishna:
Yes, I understand why MPI_[I]Rsend was used in the code, but do we have an advantage (when considering potential bugs) compared to using MPI_[I]sends? The defaults should be the more reliable of the two.

jayeshkrishna:
@jedwards4b : If you don't agree with this change we can cancel this PR. I will add flags in ACME makefile to switch to isends.

rljacob:
Its not "one buggy machine".  There is a history of this particular API being buggy especially Cray machines and there's a lot of those in our future.

jedwards4b:
The argument that one vendor can't get this API right and so we need to accommodate them is a particularly poor one - has anyone opened a ticket with cray?    But if you insist on doing it this way go ahead.

gold2718:
@jedwards4b. it may be a poor argument but it may be the only realistic option. I have crap all over my code to compensate for PGI's (one vendor) inability to get a different API (Fortran 2003) right.

jedwards4b:
In this case, the alternative - setting the flag to switch to isend for that vendor - is easy to implement.

jayeshkrishna:
Will add custom flags in ACME makefile/config_machines to switch to isends.

Issue #388, opened 2016-12-30
Add darray tests for async cases


edhartnett:
The darray tests in test_pioc.c are only run for non-async cases. 

In general the same tests will not be used for both async and non-async. Since the number of IO and computation procs change, it's easier to just have separate tests.

edhartnett:
I am going to close this ticket. There are now a number of darray async tests, and they will continue to be developed as new features are added.

PR #1015, opened 2017-05-03
Getting async darrays working

edhartnett:
In this PR the first test case of writing with async and darrays is added to the code base.

There is still work to be done here on the test, but let's get this progress merged and I will continue to develop testing in future PRs.

Part of #89.
Part of #388.

The new test test_darray_async is only run when PIO_USE_MALLOC=On. (The bget code does not work with async).


edhartnett:
Just got some new test failures. I will fix and resubmit.

PR #1045, opened 2017-05-19
Updating number of PnetCDF requests correctly

### LINKED ISSUE #1044: The number of pnetcdf requests is calculated incorrectly
This bug can be reproduced with pioperf with 4 MPI tasks (it seems that no existing unit tests reproduce it)
1) Build latest PIO master branch with pnetcdf
2) Modify tests/performance/pioperformance.F90, line 60, change the default nframes to 6
nframes = 6
3) Build pioperf from pioperformance.F90
4) Use gensimple.pl to generate simple.dat (with 4 tasks), for example:
```
version 2001 npes 4 ndims 1 
16 
0 4
1 2 3 4 
1 4
5 6 7 8 
2 4
9 10 11 12 
3 4
13 14 15 16 
```
5) Create pioperf.nl like below:
```
&pioperf
decompfile = "./simple.dat"
pio_typenames = 'pnetcdf'
niotasks = 2
rearrangers = 2
nvars = 1
/
```
6) Run pioperf with 4 MPI tasks
mpiexec -n 4 ./pioperf

The performance test will fail with MPI_Abort.

With some debugging work on vdesc->request array and vdesc->nreqs, it is observed that:
t1: Array capacity = 16, nreqs = 1, expected nreqs = 1
t2: Array capacity = 16, nreqs = 3, expected nreqs = 2
t3: Array capacity = 16, nreqs = 6, expected nreqs = 3
t4: Array capacity = 16, nreqs = 10, expected nreqs = 4
t5: Array capacity = 16, nreqs = 15, expected nreqs = 5
t6: Array capacity = 16, nreqs = 21, expeced nreqs = 6
We can see that at nframe 6, nreqs are larger than the array capacity, this will cause some pnetcdf requests to have undefined values.

The bug is from pio_darray_int.c, line 313 (original introduced by commit f397173 in Nov 25, 2015):
`vdesc->nreqs += reqn + 1;`

It should be changed to
`vdesc->nreqs = reqn + 1;`

dqwu:
Fixes #1044. 

### DIFF
--- a/src/clib/pio_darray_int.c
+++ b/src/clib/pio_darray_int.c
@@ -284,9 +284,7 @@ int write_darray_multi_par(file_desc_t *file, int nvars, int fndims, const int *
                         /* Get a pointer to the data. */
                         bufptr = (void *)((char *)iobuf + nv * iodesc->mpitype_size * llen);
 
-                        /* ??? */
-                        int reqn = 0;
-                        if (vdesc->nreqs % PIO_REQUEST_ALLOC_CHUNK == 0 )
+                        if (vdesc->nreqs % PIO_REQUEST_ALLOC_CHUNK == 0)
                         {
                             if (!(vdesc->request = realloc(vdesc->request, sizeof(int) *
                                                            (vdesc->nreqs + PIO_REQUEST_ALLOC_CHUNK))))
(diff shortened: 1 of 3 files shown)


dqwu:
@edhartnett 
Fortran unit test pio_decomp_frame_tests has been updated to reproduce the bug (should be tested with AddressSanitizer, as heap-buffer-overflow does not always cause segmentation faults)

edhartnett:
Thanks for the test!! :-)

PR #1047, opened 2017-06-06
Renaming XML man pages to avoid colons

jayeshkrishna:
Colons in file names (e.g. XML::Lite.3) creates issues when
downloading the source on a Windows machine. 

Replacing the colons with underscores.


jayeshkrishna:
At least one user in the ACME land team has issues downloading the source on his Windows machine due to this issue.

jayeshkrishna:
I have never used them :)

jayeshkrishna:
FYI: These are man pages for Lite.pm. Changing the name of the files shouldn't cause any issues.

Issue #1064, opened 2017-06-27
TARGET_LINK_LIBRARIES


mt5555:
Here is a suggestion based on my limited understanding of cmake - this might not be 100% correct:

How should applications which link with PIO construct the proper link link for the libraries pio depends on (such as netcdf)?

PIO's cmake build system spends some time constructing the proper NETCDF and PNETCDF compile and link settings.  The link settings in particular should then be set via

TARGET_LINK_LIBRARIES(pio ${NetCDF_Fortran_LIBRARIES})
TARGET_LINK_LIBRARIES(pio ${PnetCDF_Fortran_LIBRARIES})

Then other codes that use cmake and link with PIO will get the proper libraries added to their link line automatically and they dont have to repeat the netcdf/pnetcdf configure exercise.   This inheritance is a feature of cmake.  

And thus, for example, the "basic" example code included with PIO doesn't need to add the NETCDF_Fortran_Library in it's TARGET_LINK_LIBRARIES command:
TARGET_LINK_LIBRARIES(testpio piof pioc ${NETCDF_Fortran_LIBRARY} ${TIMING_LINK_LIB})

In PIO2, I cant seem to find any TARGET_LINK_LIBRARIES  calls, so I think they might be good to add.  

In PIO1, there is a bug, that the code reads:

if(PNETCDF_LIBRARY)
  TARGET_LINK_LIBRARIES(pio ${PNETCDF_LIBRARIES})
ENDIF()
if(NETCDF_LIBRARIES)
  TARGET_LINK_LIBRARIES(pio ${NETCDF_LIBRARIES})
ENDIF()

But these all-caps variables are not set by PIO's FindNetCDF.  Instead, they should use the mixed case variable names shown above.  Also, I dont think the if() is needed, since if they are not set TARGET_LINK_LIBRARIES will ignore them?

I plan on submitting a PR for pio1 to fix this, but thought I'd describe the issue first.  Since the TARGET_LINK_LIBRARIES has been removed going from pio1 to pio2, maybe pio2 has an different way to accomplish this?

PR #1067, opened 2017-06-28
fix typos in TARGET_LINK_LIBRARIES

mt5555:
Fixes issue #1064


pio/CMakeLists.txt
```diff
@@ -175,12 +175,8 @@ if("${CMAKE_Fortran_COMPILER_ID}" STREQUAL "GNU")
 endif()
 
 ADD_LIBRARY(pio ${SRCS_F90} ${SRCS_C} ${TEMPSRCF90})
-if(PNETCDF_LIBRARY)
-  TARGET_LINK_LIBRARIES(pio ${PNETCDF_LIBRARIES})
-ENDIF()
-if(NETCDF_LIBRARIES)
-  TARGET_LINK_LIBRARIES(pio ${NETCDF_LIBRARIES})
-ENDIF()
+TARGET_LINK_LIBRARIES(pio ${PnetCDF_Fortran_LIBRARIES})
```
jayeshkrishna: Can you add the conditionals (if(PnetCDF_Fortran_FOUND) and if(NetCDF_Fortran_FOUND) should work I think) back? PIO1 can be built without netcdf or pnetcdf libs (although i don't try it a binary format is supported by pio1).

jayeshkrishna:
@jedwards4b : The conditionals are still required since users can built with different combinations of netcdf, pnetcdf libs (and neither netcdf or pnetcdf)

jedwards4b:
According to MT ```Also, I dont think the if() is needed, since if they are not set TARGET_LINK_LIBRARIES will ignore them?```  I assume he tested that?

jayeshkrishna:
ok, cmake magic :)

mt5555:
Yes, I verified the if() is not needed.  

if findPnetCDF.cmake doens't find pnetcdf, the variable is unset and TARGET_LINK_LIBRARIES is smart enough to ignore it.

Issue #1037, opened 2017-05-12
Mixing types in iodesc causes wrong fill values to be used?


edhartnett:
It's possible to mix different types in the same decomposition, as long as the types are of the same size.

But if we do this, I believe the code below (from write_darray_multi_par()) will use the fill values for the first variable's type:

```
    /* Point to var description scruct for first var. */
    vdesc = file->varlist + varids[0];

    /* Set these differently for data and fill writing. */
    int num_regions = fill ? iodesc->maxfillregions: iodesc->maxregions;
    io_region *region = fill ? iodesc->fillregion : iodesc->firstregion;
    PIO_Offset llen = fill ? iodesc->holegridsize : iodesc->llen;
    void *iobuf = fill ? vdesc->fillbuf : vdesc->iobuf;

```
We need a test to check this.

edhartnett:
I am about to put up a PR which includes new test, test_darray_multivar3.c. In this test I create a file with two ints and one float, and use the same decomposition. The decomposition has holes, and the SUBSET rearranger is used.

This demonstrates the problem with fill values. The resulting file shows that when fill values are not provided, the default fill value of NC_INT is used for the float.

```
netcdf data_test_darray_multivar3_iotype_2_use_fv_0 {
dimensions:
	timestep = UNLIMITED ; // (1 currently)
	x = 4 ;
	y = 4 ;
variables:
	int Kirk(x, y) ;
	int Spock(timestep, x, y) ;
	float McCoy(timestep, x, y) ;
data:

 Kirk =
  0, 1, 2, _,
  10, 11, 12, _,
  20, 21, 22, _,
  30, 31, 32, _ ;

 Spock =
  0, 1, 2, _,
  10, 11, 12, _,
  20, 21, 22, _,
  30, 31, 32, _ ;

 McCoy =
  0.5, 1.5, 2.5, -1.401298e-45,
  10.5, 11.5, 12.5, -1.401298e-45,
  20.5, 21.5, 22.5, -1.401298e-45,
  30.5, 31.5, 32.5, -1.401298e-45 ;
}

```

An additional, possibly related problem occurs with pnetcdf files. In that case, the float data values are corrupted, in addition to the fill values being incorrect. This problem occurs even if the fill values are provided.

```
netcdf data_test_darray_multivar3_iotype_1_use_fv_1 {
dimensions:
	timestep = UNLIMITED ; // (1 currently)
	x = 4 ;
	y = 4 ;
variables:
[...]

PR #1091, opened 2017-07-28
Added test which demonstrates some problems with fill values and mixed types in iodesc

edhartnett:
See #1037 for discussion.

This PR adds a test that demonstrates problem. Code in test that causes error is commented out so that CI can continue to pass.

I have merged with develop for testing. Current uses of PIO should be worried about this bug, I think. If you mix types in your iodesc, you may have problems in your output data.

Short-term fix is to disable writing in an iodesc defined for another type.

Long-term fix is to move the type information out of the decomposition, which would be type-neutral. The code needs to be modified to use the correct fill values regardless of the decomposition. But this is a fair amount of work.

### DIFF
--- a/src/clib/pio_darray.c
+++ b/src/clib/pio_darray.c
@@ -526,8 +526,6 @@ int PIOc_write_darray(int ncid, int varid, int ioid, PIO_Offset arraylen, void *
         LOG((3, "allocating multi-buffer"));
         if (!(wmb->next = bget((bufsize)sizeof(wmulti_buffer))))
             return pio_err(ios, file, PIO_ENOMEM, __FILE__, __LINE__);
-        /* if (!(wmb->next = calloc(1, sizeof(wmulti_buffer)))) */
-        /*     return pio_err(ios, file, PIO_ENOMEM, __FILE__, __LINE__); */
         LOG((3, "allocated multi-buffer"));
 
         /* Set pointer to newly allocated buffer and initialize.*/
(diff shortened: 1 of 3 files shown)


dqwu:
Question 1: The new test file test_darray_multivar3.c is not included in this PR?
Question 2: What does the 2nd commit (added test for fill value bug when mixing types in iodesc) in this PR do? It modifies pio_darray.c, but actually nothing is changed (only a comment is deleted, and three lines are split into two lines each).

edhartnett:
Whoops! Added the missing file.

dqwu:
Should the 2nd commit (added test for fill value bug when mixing types in iodesc) be removed? Otherwise it looks misleading (it did not add a test, and it modified pio_darray.c without actual/effect code changes).

edhartnett:
I wished to draw attention to a bug which may be affecting any applications you have that use PIO.

If you are using PIO in such a way as to activate this bug, that would be a serious problem. Output data would be silently corrupted.

You will activate this bug if you are mixing types in an iodesc, and leaving gaps in your decomposition. Non-pnetcdf output in this case will still be OK if you are specifying a fill value. pnetcdf output seems to have problems whether or not you specify a fill value. (These are preliminary conclusions, based on what I know of the bug, which is not much at the moment.)

If you are confident that your applications never activate this bug, that's good. Just avoid mixing types in iodesc for a little while longer, until I code up the changes that Jim and I are discussing. In the short term, I may add error checking to reject mixing types in an iodesc.

If you direct your attention to the test code, you will see an example of how the bug may be activated.

Sorry to have confused with excess changes. I will remove them and submit them as a separate PR. This may make it easier for you to see what activates the bug, which is in the test program. Probably you want to check that your applications are not doing anything like this. ;-)

dqwu:
@edhartnett 
This new test seems to have some memory leaks, see
http://my.cdash.org/testDetails.php?test=33399540&build=1260290

Please try to fix the leaks before merging to master, thanks.

Issue #1136, opened 2017-08-09
Handle fixed-length time dimensions


dqwu:
In some existing input files of ESMCI/cime, the time dimension might have a fixed length. For example:
```
lat = 94 ;
lon = 192 ;
time = 12 ;
...
float prc(time, lat, lon) ;
```
Note: by definition, this kind of time dimension is NOT an unlimited dimension (a.k.a. record dimension). To support it, PR #1131 has removed rec_var check from setframe/advanceframe. This allows a frame number to be set even on non-record variables (with a fixed-length time dimension):
```
    /* Set the record dimension value for this variable. This will be
     * used by the write_darray functions. */
    vdesc->record = frame;
```
In PR #1088, ALL occurrences of "vdesc->record >= 0" have been replaced by "vdesc->rec_var".
However, for non-record variables with a fixed-length time dimension, it is possible that "vdesc->record >= 0" is true while "vdesc->rec_var" is false. With latest PIO2 master branch, some ESMCI/cime tests still fail due to this issue (confirmed).

For the write_darray functions, they should support writing both record variables (with an unlimited time dimension) and non-record variables (with a fixed-length time dimension). We should probably revert testing of "vdesc->rec_var" back to "vdesc->record >= 0" (partially revert PR #1088).

Issue #1128, opened 2017-08-07
Strong dimension-checking for pio_setframe broke some ESMCI/cime tests


dqwu:
For a long time, the Fortran interface pio_setframe just sets the record dimension value for a variable, without checking the dimensions.

In PR #1084, a strong dimension-checking logic was added. Only when the first dimension is unlimited, the rec_var flag is set to true for a variable. If rec_var is false, PIO_EINVAL will be returned from 
PIOc_setframe().

However, this enhancement has caused existing ESMCI/cime tests/cases to fail.

1) Failed X case:
cime code calls pio_setframe on a variable (to be written to an output file) that does not have time dimensions, and passed -1 as the frame number.
This bug could be fixed on cime side: when there is no time sample, pio_setframe should not be called.

2) Failed cime_developer tests:
cime code calls pio_setframe on a variable (read from an input file) that has a time dimension, and the frame number is also valid.
The issue is, the time dimension is not an unlimited one, which will prevent PIO from setting the rec_var flag to true.

For example, in inputdata/atm/datm7/NYF/nyf.gxgxs.T62.051007.nc 
```
lat = 94 ;
lon = 192 ;
time = 12 ;
...
float prc(time, lat, lon) ;
        prc:short_name = "prc" ;
        prc:long_name = "Precipitation" ;
        prc:units = "mm/day" ;
```
The new dimension-checking logic requires that the time dimension be unlimited, like:
time = UNLIMITED ; // (12 currently)

[Question]
For ACME/CESM, since there are many existing legacy input files like nyf.gxgxs.T62.051007.nc, should we use a weak dimension-checking for pio_setframe such that the time dimension does not have to be unlimited?

jedwards4b:
These are cime issues and an issue should be opened in that repo.  I don't think that it will be overly difficult to fix the issue on that end.

edhartnett:
I have no objection to rolling back the checking of dimensions in setframe() if that is helpful. I can do that today, either as a temporary or permanent solution.

In case 1, where setframe is being called with a -1, I agree with Jim that this is best fixed in the caller.

In case 2, are you then doing darray calls on a var without an unlimited dimension, but treating one of the dimensions as a record dimension? Does that work?

dqwu:
@jedwards4b 
On cime side, case 1 can be easily fixed.
For case 2, it could not change existing legacy input files where the time dimension in those files is not unlimited. The fix should be on PIO side, thoughts?

[...]

PR #1137, opened 2017-08-09
Add a new Fortran test for limited time dimension

dqwu:
This test case is used for issues #1128 and #1136.

I will merge it to develop branch for testing.
Note: This test can pass before PR #1084 was merged. Before issue #1136 is fixed, it should fail as expected.


jayeshkrishna:
This is ok for now, but to just test pio_setframe() on a variable with no unlimited dims copying the entire test seems to be an overkill.

dqwu:
@jayeshkrishna 
Could be an overkill. The original test provides data comparison, I think it will be useful for the new test.

jayeshkrishna:
Yes, the data comparison is important but you don't need a 4d test (a simple 2d test should work).

Issue #1135, opened 2017-08-08
Add autoconf based build system for the C library


edhartnett:
To support netCDF integration, an autotools build system is preferred. Only the C library need be buildable with the autotools build system.

PR #1138, opened 2017-08-10
Adding basic autotools build in parallel with CMake build

edhartnett:
In this PR I put the basic autotools build system in place. (Currently it builds all tests but only runs the first. More to come in future PRs.)

The autotools build system coexists with the CMake build system. In order to make this work, I had to move the config.h file from /src/clib into the root build directory. Also all tests (in addition to all code files) must now include config.h BEFORE any other includes. It must be the first include.

Part of #1135.


dqwu:
@edhartnett 
With latest develop branch, there are build errors on anlworkstation. Please take a look, thanks.
```
src/clib/pio_file.c:5:20: fatal error: config.h: No such file or directory
 #include <config.h>
                    ^
compilation terminated.
```

dqwu:
@edhartnett 
One workaround is to use "#include <../../config.h>" in code files and test files.
Is there a better solution for the building process to pass?

edhartnett:
Yes, there is a better way. Stand by while I get my next PR up and I will swing back around to this one and fix the issue.

edhartnett:
@dqwu I just commit a fix for this problem. Try again and let me know if it still doesn't work. Now working on a clean clone for me.

dqwu:
@edhartnett 
This fix works, no build errors for CMake.

PR #1150, opened 2017-08-15
added detection of netCDF-4 parallel in autotools build

edhartnett:
Also sorted out a test issue.

Part of #1135.


configure.ac
```diff
@@ -31,6 +30,19 @@ AC_DEFINE([USE_PNETCDF_VARN_ON_READ], [1], [defined by CMake build])
 AC_CHECK_LIB([netcdf], [nc_create], [], [AC_MSG_ERROR([Can't find or link to the netcdf library.])])
 AC_CHECK_LIB([pnetcdf], [ncmpi_create], [], [AC_MSG_ERROR([Can't find or link to the parallel-netcdf library.])])
 
+# Do we have a parallel build of netCDF-4?
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM([#include "netcdf_meta.h"],
```
jedwards4b: Is this a test that we can depend on independent of netcdf version?

edhartnett: Yes, for all recent versions. Actually the CMake build uses the same method.

PR #1170, opened 2017-08-25
now checking fill values for PIOc_write_darray()

### LINKED ISSUE #768: PIOc_write_darray() fill value handling uses default fill values which user may have overridden
In PIOc_write_darray() there is this code:

```
            vtype = (MPI_Datatype)iodesc->basetype;
            LOG((3, "caller did not provide fill value vtype = %d", vtype));
            if (vtype == MPI_INT)
            {
                int fill = PIO_FILL_INT;
                memcpy((char *)wmb->fillvalue + tsize * wmb->validvars, &fill, tsize);
            }
            else if (vtype == MPI_FLOAT)
            {
                float fill = PIO_FILL_FLOAT;
                memcpy((char *)wmb->fillvalue + tsize * wmb->validvars, &fill, tsize);
            }
            else if (vtype == MPI_DOUBLE)
            {
                double fill = PIO_FILL_DOUBLE;
                memcpy((char *)wmb->fillvalue + tsize * wmb->validvars, &fill, tsize);
            }
            else if (vtype == MPI_CHARACTER)
            {
                char fill = PIO_FILL_CHAR;
                memcpy((char *)wmb->fillvalue + tsize * wmb->validvars, &fill, tsize);
            }
            else
                return pio_err(ios, file, PIO_EBADTYPE, __FILE__, __LINE__);

```

There are two problems with this code. It does not handle all types, and it assumes that the default fill value should be used. In fact, the user may have set a _FillValue att for a variable, and then that fill value should be used.

### LINKED ISSUE #1144: In PIOc_write_darray(), fill value determination code should be moved to function
The function PIOc_write_darray() is very long. The code that determines the fill value if needed can easily be moved to another function.

edhartnett:
Part of #83.
Fixes #1144.
Fixes #768.

This may cause other user applications to fail if they have been using an incorrect fill value with PIOc_write_darray().


dqwu:
@edhartnett 
6 cime_developer tests failed with latest PIO develop branch, see
http://my.cdash.org/viewTest.php?onlyfailed&buildid=1276838

I have confirmed that the test failure is caused by this PR (latest master at hash d514507 works)
Please do not merge this PR to master before we fix the failed cime_developer tests.

edhartnett:
What you are possibly doing in those tests is using a fill value in PIOc_write_darray() that you did not set as a fill value for the variable.

This would be a very serious bug if you did it in a real application. You would be writing fill values that are not documented as fill values. Any calculations made from such data could be significantly corrupted, because whatever value you are using for fill value would be treated as a normal data point. Since fill values are usually extremely large or small numbers, this could have a significant effect.

So please fix your tests right away so I can get these important changes merged. Once merged, future applications will not be able to exhibit this bug.

The fix for you is to create a _FillValue attribute with the fill value that you are using in write_darray(). In fortran you do this with pio_put_att(), in C you could do the same, or use PIOc_def_var_fill() instead, which is a convenience function that will set the _FillValue attribute for you.

In this PR you can see some fixes for such tests, in both Fortran and C. So it should be easy to fix your tests.

dqwu:
@edhartnett 
These tests are not my own tests, they are existing Fortran tests in ESMCI/cime's developer tests suite. ESMCI/cime is used by both CESM and ACME, so it is important to not break the existing tests.

@jedwards4b 
ESMCI/cime uses PIO1 by default at this time, but it will use PIO2 in the future. Do you think if the failed cime_developer tests (tested with latest PIO2 develop branch) should be updated on ESMCI/cime side to meet the new requirements in this PR?
PS, even if we fix these tests, we are not sure if any CESM/ACME code/cases/tests need to be updated as well.

jedwards4b:
I've reproduced the problem but I'm not sure I understand it yet, will get back to you.

jedwards4b:
So far in cesm I see some variables that pass a fill value to pio_darray but the don't actually use it.
I want to fix it by adding a call to pio_def_var_fill but there doesn't seem to be a fortran interface to this function?

dqwu:
@jedwards4b 
Please also check at least one failed cime_developer test (e.g. DAE.f19_f19) with latest PIO2 develop branch, as they are independent of CESM or ACME.

edhartnett:
@jedwards4b There is no pio_def_var_fill() in fortran, however you can use something like:

`ierr = PIO_put_att(pio_file, pio_var, '_FillValue', BUF_FILLVAL)`

jedwards4b:
See https://github.com/ESMCI/cime/pull/1842

dqwu:
@jedwards4b 
I tested that ESMCI/cime PR with latest PIO2 develop branch, the failed 6 cime_develop tests can pass now.

edhartnett:
Can I merge this yet? Thanks!

dqwu:
@edhartnett 
ESMCI/cime#1842 has been merged so you can merge this PR now.

Issue #1134, opened 2017-08-08
Add test of init_async with more than 1 computational component


edhartnett:
Part of #403.

edhartnett:
Main problem in existing PIO async code is that the union_comm needs to be created on all processes, not just those in_cmp || in_io.

Also the tag argument on the intercom creation call is wrong. Investigation and testing continues...

PR #1172, opened 2017-08-30
added test program

edhartnett:
Part of #1134.

Test with just MPI code to establish communicators between one IO and 2 computation components.


dqwu:
@edhartnett 
Some memory leaks from test_async_mpi:
http://my.cdash.org/testDetails.php?test=34360114&build=1278746

Please check if the leaks can be fixed.

dqwu:
@edhartnett 
Valgrind build provides more detailed information:
http://my.cdash.org/testDetails.php?test=34361433&build=1278789

edhartnett:
@dqwu thanks for noticing! I also saw those fails on my own jenkins server. ;-) I have fixed leaks.

PR #1174, opened 2017-08-30
A new Fortran test to check whether pio_read_darray supports data conversion

dqwu:
This Fortran test can reproduce the issue described in #1173.

I will merge it to develop branch for testing, which is expected to pass for PnetCDF type and fail for NetCDF type.

PR #1175, opened 2017-08-30
Adding data conversion support for NetCDF type

### LINKED ISSUE #1173: No data conversion in pio_read_darray for NetCDF type
Each NetCDF variable has an external type, specified when the variable is first defined. For nc_get_vara (called by pio_read_darray), the type of the data in memory must match the external type of the variable - no data conversion is done. We should use the members of the nc_get_vara_type family instead - if the type of data value differs from the NetCDF variable type, type conversion will occur.

ACME test ERS.f09_g16.I1850CLM45CN fails with PIO2 due to this issue. When reading a variable, the IO type is PIO_IOTYPE_NETCDF, the data buffer is double type, but the variable is float type (defined in the nc file). The data buffer will get incorrect double type values.

The following example code shows that without data conversion we get unexpected values. 
```
/*
  [    D1  ][   D2   ]
  [ F1  F2 ][        ]
*/

#include <stdio.h>
#include <string.h>

int main()
{
  double data[2];
  float f1 = 1.1;
  float f2 = 2.2;
  float* pf = (float*)data;
  double* pd = data;

  memset(data, 0, sizeof(data));
  pf[0] = f1;
  pf[1] = f2;
  printf("Without data conversion, data[0] = %lf, data[1] = %lf\n", data[0], data[1]);

  memset(data, 0, sizeof(data));
  pd[0] = f1;
  pd[1] = f2;
  printf("With data conversion, data[0] = %lf, data[1] = %lf\n", data[0], data[1]);

  return 0;
}
```
The output is:
Without data conversion, data[0] = 3.600001, data[1] = 0.000000
With data conversion, data[0] = 1.100000, data[1] = 2.200000

PIO1 does not have this issue as the Fortran APIs used by pio_read_darray support data conversion.

PnetCDF type also does not have this issue as ncmpi_get_varn_all is used by PIO2. For that API, if the type of data values differs from the NetCDF variable type, type conversion will occur. 

I will add a new Fortran test for PIO2 to reproduce this issue.

dqwu:
Fixes #1173 

The new Fortran test in PR #1174 should pass with this fix.

### DIFF
--- a/src/clib/pio_darray_int.c
+++ b/src/clib/pio_darray_int.c
@@ -867,7 +867,48 @@ int pio_read_darray_nc(file_desc_t *file, io_desc_t *iodesc, int vid, void *iobu
             {
 #ifdef _NETCDF4
             case PIO_IOTYPE_NETCDF4P:
-                ierr = nc_get_vara(file->fh, vid, start, count, bufptr);
+                /* ierr = nc_get_vara(file->fh, vid, start, count, bufptr); */
+                switch (iodesc->piotype)
+                {
+                case PIO_BYTE:
+                    ierr = nc_get_vara_schar(file->fh, vid, start, count, (signed char*)bufptr);
+                    break;
+                case PIO_CHAR:
+                    ierr = nc_get_vara_text(file->fh, vid, start, count, (char*)bufptr);
+                    break;
+                case PIO_SHORT:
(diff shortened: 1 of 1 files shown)


edhartnett:
Good catch, thanks!

edhartnett:
I am going to merge this PR, since it definitely fixes a bug.

PR #1198, opened 2017-09-13
Fix a random failure (on MPI_Type_free) for test_rearr 

### LINKED ISSUE #1197: test_rearr fails randomly on MPI_Type_free
ios.compproc is not initialized in the function below (its value is random):
```
/* Test define_iodesc_datatypes() function. */
int test_define_iodesc_datatypes()
{
    ...
    /* Run the functon. */
    for (int r = 0; r < NUM_REARRANGERS; r++)
    {
        iosystem_desc_t ios;
        io_desc_t iodesc;

        /* Set up test for IO task with BOX rearranger to create one type. */
        ios.ioproc = 1; /* this is IO proc. */
        ios.num_iotasks = 4; /* The number of IO tasks. */
        iodesc.rtype = NULL; /* Array of MPI types will be created here. */
        ...
        if ((ret = define_iodesc_datatypes(&ios, &iodesc)))
            return ret;
        ...
    }
    ...
}
```
Note that define_iodesc_datatypes() function checks whether compproc is 0 or not. When compproc happens to be 0 , this test will fail on MPI_Type_free:
```
int define_iodesc_datatypes(iosystem_desc_t *ios, io_desc_t *iodesc)
{
    ...
    if (ios->compproc)
    ...
}
```
For this test, each task is a compute task so we should set ios.compproc to 1.

dqwu:
Fixes #1197 

I will merge this fix to the develop branch for testing.

### DIFF
--- a/tests/cunit/test_rearr.c
+++ b/tests/cunit/test_rearr.c
@@ -494,6 +494,7 @@ int test_define_iodesc_datatypes()
 
         /* Set up test for IO task with BOX rearranger to create one type. */
         ios.ioproc = 1; /* this is IO proc. */
+        ios.compproc = 1;
         ios.num_iotasks = 4; /* The number of IO tasks. */
         iodesc.rtype = NULL; /* Array of MPI types will be created here. */
         iodesc.nrecvs = 1; /* Number of types created. */


edhartnett:
Thanks, good catch.

PR #1231, opened 2017-10-18
Partial fix for incorrect type of 'mpi_type' in pioc_support.c

jedwards4b:
The type for mpi_type in the routines inq_file_metadata and PIOc_openfile_retry
in pioc_support.c is int but should be MPI_Datatype because mpi_type is used in
calls to, e.g., MPI_Type_size (https://www.open-mpi.org/doc/v2.0/man3/MPI_Type_size.3.php).
In most implementations of MPI, the size of MPI_Datatype is the same as the size
of int, but in some versions of OpenMPI, MPI_Datatype actually has a size of 8.
This leads to memory errors and segfaults in PIO.

### DIFF
--- a/src/clib/pioc_support.c
+++ b/src/clib/pioc_support.c
@@ -1937,7 +1937,7 @@ int check_unlim_use(int ncid)
  * @author Ed Hartnett
  */
 int inq_file_metadata(file_desc_t *file, int ncid, int iotype, int *nvars, int **rec_var,
-                      int **pio_type, int **pio_type_size, int **mpi_type, int **mpi_type_size)
+                      int **pio_type, int **pio_type_size, MPI_Datatype **mpi_type, int **mpi_type_size)
 {
     int nunlimdims;        /* The number of unlimited dimensions. */
     int unlimdimid;
@@ -1966,7 +1966,7 @@ int inq_file_metadata(file_desc_t *file, int ncid, int iotype, int *nvars, int *
             return PIO_ENOMEM;
         if (!(*pio_type_size = malloc(*nvars * sizeof(int))))
             return PIO_ENOMEM;
-        if (!(*mpi_type = malloc(*nvars * sizeof(int))))
+        if (!(*mpi_type = malloc(*nvars * sizeof(MPI_Datatype))))
             return PIO_ENOMEM;
         if (!(*mpi_type_size = malloc(*nvars * sizeof(int))))
             return PIO_ENOMEM;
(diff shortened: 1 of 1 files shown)


jedwards4b:
I need to understand why this is a partial fix?

mgduda:
@jedwards4b I referred to this as a "partial" fix in my original commit mostly because I don't think I understand fully the structure of the code in pioc_support.c well enough to judge whether there are any places that I've failed to account for the difference between `int` and `MPI_Datatype`. The essentially point, though, is simply that `sizeof(MPI_Datatype)` may not be the same as `sizeof(int)` for certain MPI implementations.

src/clib/pioc_support.c
```diff
@@ -2368,7 +2368,7 @@ int PIOc_openfile_retry(int iosysid, int *ncidp, int *iotype, const char *filena
             return check_mpi(file, mpierr, __FILE__, __LINE__);
         if ((mpierr = MPI_Bcast(pio_type_size, nvars, MPI_INT, ios->ioroot, ios->my_comm)))
             return check_mpi(file, mpierr, __FILE__, __LINE__);
-        if ((mpierr = MPI_Bcast(mpi_type, nvars, MPI_INT, ios->ioroot, ios->my_comm)))
+        if ((mpierr = MPI_Bcast(mpi_type, nvars*(int)(sizeof(MPI_Datatype)/sizeof(int)), MPI_INT, ios->ioroot, ios->my_comm)))
```
edhartnett: I believe this should be:
MPI_Bcast(mpi_type, nvars, MPI_Datatype, ios->ioroot, ios->my_comm);

mgduda: The third argument to `MPI_Bcast` should be the size of the elements in the buffer that is the first argument. As far as I know, there isn't an MPI datatype definition for `MPI_Datatype` (that is, `MPI_INT` is use for `int`, `MPI_FLOAT` is used for `float`, etc., but I don't know of anything that is used for `MPI_Datatype`). My solution was to treat the size of the buffer as `MPI_INT`, but to set the size of the buffer as the normal size scaled by the ration of `sizeof(MPI_Datatype)/sizeof(int)`.

jedwards4b:
I agree with @mgduda and have merged to develop for testing

src/clib/pioc_support.c
edhartnett: OK, I get what you are saying but this is a rather convoluted and non-obvious way to handle it.

I think a better way would be to cast the value to long int and use the MPI_LONG_INT type.

jedwards4b: That would require an if block - maybe we just need some documentation at this line explaining the need for convolution?

mgduda:
@jedwards4b How is the testing going? Is there anything I can do to help out?

PR #1232, opened 2017-10-27
Norequests

jedwards4b:
Git rid of all of the request management and let pnetcdf handle it.  Requires pnetcdf 1.8.0 or newer.


jedwards4b:
It turns out that the bug I was chasing that caused me to make this change was actually in pnetcdf 
(version 1.8.1 and older nonblocking.c)  so I am open to a discussion regarding whether we should keep this PR or reject it.   A new pnetcdf version is in the planning stages.

jedwards4b:
Closing without merging, but we might consider this change at a later date.

PR #1234, opened 2017-11-22
remove usage of NC_MAX_DIMS

jedwards4b:
Remove usage of NC_MAX_DIMS which is a very large number in the pnetcdf 1.9.0 release candidate.  I have merged to develop for testing.


jedwards4b:
There is also a problem in pio_msg.c @edhartnett which involves the async interface @edhartnett if dimids is present is it clear that ndims is as well?

jedwards4b:
I've redone this in dontuse_nc_max paying closer attention to testing.

Issue #1246, opened 2017-12-14
BOX Rearranger fails to distribute io


jedwards4b:
(Wu, Danqing)
[test_box.c.txt](https://github.com/NCAR/ParallelIO/files/1560530/test_box.c.txt)
This issue was observed on Mira, and I have created a simple C test program (serial, no MPI required, with the same BOX rearranger code) to reproduce it, see attached file.

In the test, we have:
    num_iotasks = 112;
    ndims = 2;
    gdimlen[0] = 288;
    gdimlen[1] = 777602;

The result of the BOX rearrangement (use DEFAULT_BLOCKSIZE 1024)
{
io_rank = 0, num_aiotasks = 1, start[0] = 0, start[1] = 0, count[0] = 288, count[1] = 777602
io_rank = 1, num_aiotasks = 1, start[0] = 0, start[1] = 0, count[0] = 0, count[1] = 0
...
io_rank = 111, num_aiotasks = 1, start[0] = 0, start[1] = 0, count[0] = 0, count[1] = 0
}

Using another global size (gdimlen[0] = 72, gdimlen[1] = 1382400) can also reproduce the same issue:
{
io_rank = 0, num_aiotasks = 1, start[0] = 0, start[1] = 0, count[0] = 72, count[1] = 1382400
io_rank = 1, num_aiotasks = 1, start[0] = 0, start[1] = 0, count[0] = 0, count[1] = 0
...
io_rank = 111, num_aiotasks = 1, start[0] = 0, start[1] = 0, count[0] = 0, count[1] = 0
}

This means it fails to converge until the number of actual IO tasks is reduced to 1. Essentially this assigns the whole global size (288 x 777602) to IO rank 0, which does not give us an evenly distribution.

In the first case, maxbytes is set to (1024 + 256 = 1280) and minbytes is set to (1024 - 256 = 768). If the data type is double (8 bytes) then the required number of procs is

(288 * 777602 * 8) / 768 = 2,332,806

This is far more than the initial number of procs (112) so use_io_procs is set to 112

    /* Find the number of ioprocs that are needed so that we have
     * blocksize data on each iotask*/
    use_io_procs = max(1, min((int)((float)pgdims / (float)minblocksize + 0.5), num_io_procs));

Then, in average, on each procs the number of bytes is (288 * 777602 * 8) / 112 = 15,996,384 ( 15.25 MB )

The issue is that maxbytes is not updated to a number a bit higher than15.25 MB.

[...]

PR #1247, opened 2017-12-14
more reasonable maxblock value

jedwards4b:
more reasonable calculation of maxblocksize.   
merged to develop
Fixes issue #1246


dqwu:
@jedwards4b 
Looks good. Could you please include the issue number in this PR for reference?

PR #1262, opened 2018-01-03
Fix fill value handling to match current netcdf behavior

### LINKED ISSUE #1258: Autotools test script returns 0 even when there are test failures
test_pioc is actually failing, but make check is returning 0.

### LINKED ISSUE #1259: test_pioc failing with latest version of netCDF
Test fails with a recent version of netCDF.

edhartnett:
NetCDF fill value behavior changed a bit in the next release (4.6.0, which will probably be out this month). A bunch of bugs got fixed (by me) in fill value handling.

Furthermore, another netCDF user has contribute a fix which brings netcdf-4 fill value handling to netCDF classic files as well, which will simplify the fill value situation. (But the code in this PR will continue to work.)

Also starting to switch to new NC3/NC4 functions. I need to import a bunch of netCDF headers. In future PRs I will see what I can do about combining or in some other way dealing with these. (Perhaps by convincing the netCDF team to make them public header files.)

Jim, why don't we do a PIO release before this PR gets merged. It's the beginning of a lot of changes that deserve a fresh version.

Fixes #1259. 
Fixes #1258. 
  
FYI starting today I am using pnetcdf-1.9.0 for testing. No problems yet...


src/clib/pio_internal.h
```diff
@@ -12,6 +12,13 @@
 
 #include <config.h>
 #include <pio.h>
+#include <nc.h>
+#include <ncdispatch.h>
```
jedwards4b: Does this restrict the versions of netcdf we can use or prohibit building pio with just an installed netcdf (are these include files normally installed with the netcdf distributon?)

edhartnett: If we release 2.3.1 now, then we will call the next release 2.4.

The 2.4 release will contain whatever header files it needs. Full compatibility for all current PIOc_ applications will be maintained. It will continue to work with all versions of netCDF 4.4.1 and after. I may have to turn off some tests so they don't activate some of the fill value bugs in those releases.

However users are strongly urged to upgrade to netcdf 4.6.0 quickly as it fixes some important fill value bugs. Specifically, 0 is mistakenly used for default fill value.

edhartnett: To address your other concern, users will not have to re-build netCDF. They can use any recent version.

But they should upgrade because of the fill value bugs. Those actually affect PIO only peripherally, as PIO turns off fill mode by default. But I have some tests for PIOc_set_fill() that will probably have to be guarded in some way against old netCDF builds.

jedwards4b: The concern is not whether users will need to rebuild netcdf, but rather whether users will be able to build pio from an installed netcdf library - it looks like for example ncdispatch.h is not installed and so users would need a netcdf dev package.

edhartnett: Users will be able to build PIO with any netCDF after 4.4.1. Whatever header files are needed will be included with PIO.

edhartnett:
I am actually going to hold off on this change. I will circle back around to this after I have merged some other changes that will help me with this merge.

PR #1284, opened 2018-02-14
handle pnetcdf/netcdf4p errors better

### LINKED ISSUE #1283: ierr non-zero from nonroot io task
In parallel IO operations it is possible (but thankfully rare) that the ioroot gets a 0 return code while another iotask gets an error.   The current logic of check_netcdf will cause a model hang in this case.

jedwards4b:
Fixes #1283 
Parallel netcdf and netcdf4p may issue errors on tasks other than the ioroot, currently this would create a hang.   Also we cannot return at line 415 in any case since their are collective MPI calls later in the routine.

### DIFF
--- a/src/clib/pioc_support.c
+++ b/src/clib/pioc_support.c
@@ -410,10 +410,9 @@ int check_netcdf2(iosystem_desc_t *ios, file_desc_t *file, int status,
     /* User must provide this. */
     pioassert(fname, "code file name must be provided", __FILE__, __LINE__);
 
-    /* No harm, no foul. */
-    if (status == PIO_NOERR)
-        return PIO_NOERR;
-
+    if (file && file->iosystem->ioproc &&
+	(file->iotype == PIO_IOTYPE_PNETCDF || file->iotype == PIO_IOTYPE_NETCDF4P))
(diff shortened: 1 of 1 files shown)


jedwards4b:
Merged to develop for testing (are you still running tests this way @edhartnett?)

edhartnett:
yes let me confirm in the morning.

jcipriani:
@edhartnett and @jedwards4b  -->  my name is James Cipriani, and I work for the Weather Company.  We have been using the NCAR MPAS model for some time now (which is dependent on PIO), and I am afraid that the current merge (#1284) has resulted in MPAS crashing every time, despite a successful compile of all software and prerequisites.  When I revert back to commit ID #7c92c489015430233727169d861a5aa1534763ec (from January 11), MPAS runs fine.  I am in the process of tracing the exact point of failure within the MPAS source, but I wanted to make you both aware and ask what you think the next steps should be.  Perhaps this is more of a question (and fix) for the MPAS team at NCAR, but I'm not really sure at this stage.  Thank you for your time and consideration on this matter.

jcipriani:
@edhartnett and @jedwards4b  -->  this is MPAS version 5.2, btw.

PR #1285, opened 2018-02-15
Add pnetcdf vard2

jedwards4b:
pnetcdf has added a vard api which uses an MPI_DATATYPE as an input to describe the data.  This is implemented in cmake with PIO_USE_PNETCDF_VARD.   I've tested both with and without this feature and all tests pass on cheyenne with netcdf-mpi/4.4.1.1 and pnetcdf r3683

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,15 @@ option (PIO_TEST_BIG_ENDIAN  "Enable test to see if machine is big endian"  ON)
 option (PIO_USE_MPIIO        "Enable support for MPI-IO auto detect"        ON)
 option (PIO_USE_MPISERIAL    "Enable mpi-serial support (instead of MPI)"   OFF)
 option (PIO_USE_MALLOC       "Use native malloc (instead of bget package)"  OFF)
+option (PIO_USE_PNETCDF_VARD       "Use pnetcdf put_vard "  OFF)
 option (WITH_PNETCDF         "Require the use of PnetCDF"                   ON)
+
+# Set a variable that appears in the config.h.in file.
+if(PIO_USE_PNETCDF_VARD)
+  set(USE_VARD 1)
+else()
+  set(USE_VARD 0)
+endif()
 
 # Set a variable that appears in the config.h.in file.
 if(PIO_USE_MALLOC)
--- a/cmake_config.h.in
+++ b/cmake_config.h.in
@@ -1,4 +1,4 @@
-/** @file 
+/** @file
  *
  * This is the template for the config.h file, which is created at
  * build-time by cmake.
@@ -27,5 +27,7 @@
 
 /* buffer size for darray data. */
 #define PIO_BUFFER_SIZE @PIO_BUFFER_SIZE@
+
+#define USE_VARD @USE_VARD@
 
 #endif /* _PIO_CONFIG_ */
--- a/examples/c/darray_no_async.c
+++ b/examples/c/darray_no_async.c
@@ -28,7 +28,7 @@
 #define NDIM3 3
 
 /* The number of timesteps of data. */
-#define NUM_TIMESTEPS 1
+#define NUM_TIMESTEPS 2
 
 /* The length of our sample data in X dimension.*/
 #define DIM_LEN_X 4
--- a/examples/c/darray_no_async.c
+++ b/examples/c/darray_no_async.c
@@ -28,7 +28,7 @@
 #define NDIM3 3
 
 /* The number of timesteps of data. */
-#define NUM_TIMESTEPS 2
+#define NUM_TIMESTEPS 1
 
 /* The length of our sample data in X dimension.*/
 #define DIM_LEN_X 4
@@ -125,8 +125,8 @@ int check_file(int iosysid, int ntasks, char *filename, int iotype,
     for (int d = 0; d < NDIM3; d++)
     {
         char my_dim_name[NC_MAX_NAME];
-        PIO_Offset dimlen; 
-        
+        PIO_Offset dimlen;
+
         if ((ret = PIOc_inq_dim(ncid, d, my_dim_name, &dimlen)))
             return ret;
         if (dimlen != (d ? dim_len[d] : NUM_TIMESTEPS) || strcmp(my_dim_name, dim_name[d]))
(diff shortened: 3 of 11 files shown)


jedwards4b:
merged to develop

edhartnett:
Is it faster?

jedwards4b:
I am just starting to look at the performance, but I expect it will be faster.    Currently I am calling for one variable at a time, there is also a way to call for multiple variables - but the computation of displacements seems quite complicated and so I haven't attempted it yet.

jedwards4b:
Yes I am seeing significant performance improvements, especially in the subset rearranger results. 
These results are from the pioperf 1024 decomps, first column is with vard second without:
 
BOX 221.945248798 253.907357822 % diff = 0.144008980584
SUBSET 17.2022264498 27.2610530152 % diff = 0.584739806487
BOX 0.7970931341 15.6004142646 % diff = 18.5716329714
SUBSET 23.0762804697 23.2784120856 % diff = 0.00875928060267
BOX 16.5457560022 17.1481030928 % diff = 0.0364049300932
SUBSET 0.1767208003 23.6305299272 % diff = 132.716743513
BOX 3.6683766478 51.0189733975 % diff = 12.907779461
SUBSET 1.0480234638 101.862109979 % diff = 96.1944937275
BOX 0.7357549043 28.5790371317 % diff = 37.843148669
SUBSET 15.3969591643 32.6897335124 % diff = 1.12312919477
BOX 70.969768347 76.8909313202 % diff = 0.0834321868468
SUBSET 1.265807036 104.813372962 % diff = 81.803594846
BOX 1.8363988215 15.8898785047 % diff = 7.65273834783
SUBSET 0.495175083 52.1571324957 % diff = 104.330688652
BOX 1.4840845103 4.9129480821 % diff = 2.31042339436
SUBSET 2.9385983806 65.4756431715 % diff = 21.2812493207
BOX 20.0069687842 181.495621358 % diff = 8.07162015974
SUBSET 16.2239472059 66.5648315442 % diff = 3.10287525591
BOX 64.8644429895 17.0219776789 % diff = -0.737576137335
SUBSET 1.0180877579 89.8994202024 % diff = 87.3022308291
BOX 0.5136319511 21.3043473794 % diff = 40.4778468002
SUBSET 0.5097111875 31.2444823791 % diff = 60.2984041656
BOX 3.4229569293 86.4888338932 % diff = 24.2672866412
SUBSET 16.1064023695 47.7631882347 % diff = 1.96547839418
BOX 45.6536133697 58.4048540982 % diff = 0.279304085423
SUBSET 0.4469981755 52.1623455094 % diff = 115.694761564
BOX 3.1092585202 74.5111987228 % diff = 22.964298317
SUBSET 3.0913546425 108.693742686 % diff = 34.1605542734
BOX 253.070705125 225.509454072 % diff = -0.108907315207
SUBSET 17.3791197277 108.152562521 % diff = 5.22313236894
BOX 1350.54126799 2416.98996633 % diff = 0.789645398937
SUBSET 378.362042 985.767949655 % diff = 1.60535635246
BOX 20.488644383 0.9091068199 % diff = -0.955628747178
SUBSET 0.2720446673 30.9646412561 % diff = 112.821901247
BOX 56.1804306688 137.740222875 % diff = 1.45174736533
SUBSET 184.742634814 213.989423372 % diff = 0.158310985374
BOX 315.221148307 203.260499107 % diff = -0.355181274484
SUBSET 9.6016766154 286.956276013 % diff = 28.8860592277
BOX 122.385732979 16.3884653647 % diff = -0.866091700676
SUBSET 31.770819415 54.388099901 % diff = 0.711888484542
BOX 79.0809470867 67.7930595279 % diff = -0.142738396221
SUBSET 3.9021360781 133.008477489 % diff = 33.0860684575
BOX 362.763510788 142.544870949 % diff = -0.607058409377
SUBSET 21.0708585134 448.016980338 % diff = 20.2623980202
BOX 1.7605992719 59.2399914571 % diff = 32.6476291923
SUBSET 71.7064753857 89.1340637456 % diff = 0.243040649623
BOX 0.7863706291 19.7163512334 % diff = 24.0725936394
SUBSET 0.7996509315 31.2486100307 % diff = 38.0778135806

edhartnett:
Is the only difference the datatype? Or is there something else happening?

jedwards4b:
The only difference is the datatype.   But this was with only one run of each case.   I'm going to redo this with 4 runs each to increase confidence in results.

edhartnett:
OK perhaps we are looking up the MPI data type each time, and can cache it instead, to also get a speed up.

jedwards4b:
I am creating it in the write, we may be able to partially cache it but it depends on record number and so the final step will always need to be redone I think.

dqwu:
@jedwards4b 
ncmpi_sync call has been removed. Need test with and without this change, to see how much performance improvement is from PIO_USE_PNETCDF_VARD and how much is from removing ncmpi_sync call, respectively.

PS, we should not remove ncmpi_sync from PIOc_sync, since when client code calls PIOc_sync it is expecting a sync operation. Instead, we should modify PIOc_close to avoid calling ncmpi_sync (implicitly via PIOc_sync) before ncmpi_close.

jedwards4b:
Tests with and without vard both remove ncmpi_sync - I have run the test suite as well as several real cases and not found a case where removing it causes a failure.

jedwards4b:
I reran the above tests 4 times and on average I see significant performance improvements with the vard interface, especially the subset rearranger seems to benefit significantly.

dqwu:
@jedwards4b 
Sync will guarantee the data is safely stored on the hard disks before the function call returns. Safe means even if file servers crashed right after the function returns, data is safely stored on disks.

With this change, the issue is: 
When user/client code calls PIOc_sync (or Fortran interface pio_sync) and a sync is actually desired (incurs a huge performance penalty, though), but the API no longer calls ncmpi_sync.

Maybe ncmpi_sync should be a CMake option for users?

jedwards4b:
How about adding an optional argument to PIOc_sync/ pio_sync which indicates whether ncmpi_sync should be called?  I haven't found a case yet where its needed, but if this optional argument was available I would probably turn it on when debugging.

wkliao:
May I suggest to create a new PIO API called pio_flush?
It essentially does the same task as the current pio_sync but without calling ncmpi_sync.
pio_sync can be changed to simply call ncmpi_sync or nc_sync, and if any PIO dirty cache is detected, pio_flush is triggered first.

rljacob:
Has PIO with this vard option made it in to CIME?

jedwards4b:
no

PR #1305, opened 2018-06-22
avoid creating MPI derived datatypes whenever possible

wkliao:
This patch checks whether a request is contiguous in file. If yes, it simply uses the offset and length of the request and skips the creation of MPI derived datatype. This should reduces cost of filetype construction. Running a small decomposition file, piodecomp30tasks04dims13.dat, reveals noticeable improvement.


jedwards4b:
Superseded by #1306

PR #1311, opened 2018-08-06
Do not overload user-defined PKGCOMPs

bekozi:
Allows overload of environment variables like MPISERIAL_C_LIBRARY without it being set to NOTFOUND


jedwards4b:
Thanks

bekozi:
Nice. Thanks for the quick merge

Issue #1316, opened 2018-12-14
bug in subset decomp


jedwards4b:
The decomp file https://svn-ccsm-piodecomps.cgd.ucar.edu/trunk/72/piodecomp72tasks02dims01.dat shows an error when using 
the following pioperf.nl on 72 tasks
```
&pioperf
decompfile = '/glade/u/home/jedwards/sandboxes/piodecomps/72/piodecomp72tasks02dims01.dat'
pio_typenames = 'pnetcdf'
rearrangers = 2
nframes=1
nvars = 1
/
```

jedwards4b:
Actually I think that this is a problem in the decomp file - the dof on a given node should be monotonically increasing but is not for example on node 50.   The error in pio is that it didn't flag this as a bad input or reorder field to be monotonic.

jedwards4b:
Fixed by #1318

PR #1318, opened 2018-12-15
Trytosort

jedwards4b:
If the supplied compmap is not monotonically increasing sort it before copying to iodesc and use the sorted map for rearranging.  If sort is required it will be done on each field prior to calling rearrange routines.   I've merged this to develop for testing.  @edhartnett @jayeshkrishna I would really appreciate your feedback if/when you have time to consider this.
Thanks


jayeshkrishna:
You might also want to add C/Fortran tests for this case.

jedwards4b:
Fixes #1316

PR #1335, opened 2019-02-13
remove restrictions to data conversion

### LINKED ISSUE #1334: Internal data conversion needs to work
Currently the check https://github.com/NCAR/ParallelIO/blob/master/src/clib/pio_darray.c#L650
is preventing pio from converting datatypes internally - this is an important feature for many users of pio - (converting double to single precision before a write or single to double after a read).  Reimplement support for this important feature.

jedwards4b:
This removes the restrictions to data conversion at the netcdf/pnetcdf library level.   All tests pass but only because I removed a few expected failures.  I plan to add back some tests before merging to develop.  
Fixes #1334

### DIFF
--- a/src/clib/pio_darray.c
+++ b/src/clib/pio_darray.c
@@ -145,8 +145,8 @@ int PIOc_write_darray_multi(int ncid, const int *varids, int ioid, int nvars,
         var_desc_t *vdesc;
         if ((ierr = get_var_desc(varids[v], &file->varlist, &vdesc)))
             return pio_err(ios, file, ierr, __FILE__, __LINE__);
-        if (vdesc->pio_type != iodesc->piotype)
-            return pio_err(ios, file, PIO_EINVAL, __FILE__, __LINE__);
+//        if (vdesc->pio_type != iodesc->piotype)
+//            return pio_err(ios, file, PIO_EINVAL, __FILE__, __LINE__);
     }
 
     /* Get a pointer to the variable info for the first variable. */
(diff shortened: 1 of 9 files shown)


edhartnett:
I was just doing this very thing. ;-) You beat me to it.

src/clib/pio_darray.c
```diff
@@ -647,10 +648,10 @@ int PIOc_write_darray(int ncid, int varid, int ioid, PIO_Offset arraylen, void *
 
     /* If the type of the var doesn't match the type of the
      * decomposition, return an error. */
-    if (iodesc->piotype != vdesc->pio_type)
-        return pio_err(ios, file, PIO_EINVAL, __FILE__, __LINE__);
-    pioassert(iodesc->mpitype_size == vdesc->mpi_type_size, "wrong mpi info",
-              __FILE__, __LINE__);
```
edhartnett: Should we check that the size of the type is the same? Because if you do this with a 1-byte and a 4-byte type, won't hilarious confusion ensue?

jedwards4b: Yes - problems will occur if you try to use an iodesc defined for one type size with a variable of another type size.   I am working on a test.

jedwards4b: It's not possible as far as I can tell to capture this error - in the call 
 PIOc_write_darray(ncid, varid, ioid, arraylen, test_data, fillvalue)

the type of test_data must be of the same type_size as the type the ioid was declared with, but there is no way to test for that since the type in varid is the files type and not necessarily the type of the variable in memory.

jedwards4b:
Merged to develop for testing

jedwards4b:
I have added changes to address issue #1336 variables must match in all dimensions to use the write_multivar functions.

Issue #1339, opened 2019-02-28
Very slow IO time when writing sub-hourly data


johnmauff:
When sub-hourly data is turned on in CESM2, the write call gets progressively slower. There is what looks like a linear cost term that is based on the number of calls to PIO_write_darray. I believe it is due to the use of a single linked list in the subroutine "add_data_to_buffer" in the piodarray.F90.in file. Because each write is so small when generating sub-hourly data, the linked list file%data_list_top gets to be very long before the buffer data is actually written to disk. A reduction in the pio_buffer_size_limit appears to significantly increase IO performance because it reduces the maximum length of the linked list. Replacing a single linked list with a double linked list should resolve this issue while still maximizing the size of IO to disk.

jedwards4b:
John,
If I understand you correctly we don't really need a doubly linked list - we just need another variable in the File structure that always points to the end of the list so that you don't have to traverse the entire list whenever you want to append.

johnmauff:
Jim,

  Yes that is potentially another solution that would likely be simpler.

jedwards4b:
I've pushed a branch to my fork at https://github.com/jedwards4b/ParallelIO/
branch name pio1_linkedlist_opt 
please give it a shot.

johnmauff:
Thanks for the new code.    That reduced I/O time for this particular configuration by 40x.

rljacob:
Which system was this observed on?

jedwards4b:
@rljacob cheyenne/intel  but it's a pretty system independent optimization.

PR #1340, opened 2019-03-01
optimize the linked list with a pointer to the list end

jedwards4b:
Add a pointer to the end of the linked list to avoid traversing the list each time add_data_to_buffer_{TYPE}  is called.


jedwards4b:
Fixes issue #1339

PR #1361, opened 2019-03-18
E3SM fork changes - DO NOT MERGE - for discussion only

jayeshkrishna:
This PR contains all the current changes in the E3SM PIO fork.

Although merging this PR directly is going to be difficult (priorities, conflicts etc), it can be used as a reference PR for cherry picking commits that might be interesting to users.


jayeshkrishna:
@jedwards4b / @edhartnett ,
If needed we can have a call sometime next week to discuss the changes at a high level.

I think support for ADIOS in PIO might be interesting to you.

Merging the PR as is might be difficult.

jedwards4b:
@jayeshkrishna Thank you for posting this back to the repository I'll take a look and discuss with @edhartnett    We will probably want to discuss with you as well.

edhartnett:
Yes, super interesting. Seems you have added a C++ API?

jedwards4b:
@jayeshkrishna I'm unable to build the branch with the intel compiler - do you know what I need to do?
```
 Building CXX object src/clib/CMakeFiles/pioc.dir/pio_sdecomps_regex.cpp.o
In file included from /glade/u/apps/ch/opt/mpt/2.19/opt/hpe/hpc/mpt/mpt-2.19/include/mpi.h(1817),
                 from /glade/u/home/jedwards/sandboxes/ParallelIO.e3sm/src/clib/pio.h(16),
                 from /glade/u/home/jedwards/sandboxes/ParallelIO.e3sm/src/clib/pio_sdecomps_regex.cpp(20):
/glade/u/apps/ch/opt/mpt/2.19/opt/hpe/hpc/mpt/mpt-2.19/include/mpi++.h(416): error: more than one instance of overloaded function "PMPI::Init" has "C" linkage
  Init();
  ^
```

jayeshkrishna:
This is most likely due to the MPI headers that get included via the C++ source (Pio.h is included as a C header within the C++ source, and pio.h includes mpi.h). Can you try directly including mpi.h at the beginning of pio_sdecomp_regex.cpp (there should already be an #ifdef _ADIOS section that includes mpi.h before the extern C headers in pio_sdecomp_regex.cpp, try removing the #ifdef _ADIOS and see if it works)?
I will also try it out and see if I can reproduce the issue.
@dqwu : Have we built on Cori with intel before/after the ADIOS merge?

edhartnett:
What do you think about making the PIO C++ library a separate repo, with it's own build system(s). Instead of being part of the PIO C/Fortran package, it will depend on the PIO C library being installed.

What we learned with netCDF was putting all your APIs together in one project seems to make sense, but ends of leading to complex build systems and users having support issues with language APIs they were not planning on using. (As Jim is now having C++ API build issues, though he only wants the C/Fortran APIs.)

edhartnett:
Also can you tell us more about the ADIOS iotype?

jayeshkrishna:
The ADIOS iotype supports writes using the ADIOS library. It might make more sense discussing the details over a conference call.

rljacob:
https://csmd.ornl.gov/adios

edhartnett:
Ok, let's back up and take a larger view. We have already identified two major features in this PR:
* C++ API
* support for ADIOS.

Are there more major features? What else is in this PR? ;-)

jayeshkrishna:
I just tried Intel compiler (18) + Intel MPI on Cori with our fork master and did not see the issue (MPI header issue - including from C++). I remember the ADIOS developers mentioning that they encountered a similar issue, so I will check with them too.

jayeshkrishna:
I will include a comment soon, charting out all the changes in the PR at a high level.

dqwu:
I would also recommend two changes:
* Optimize the box rearrange create algorithm.
This greatly improves performance of BOX rearranger.
* A new version of GCDblocksize()
Note, the old version has a bug when it is converted from PIO1 Fortran code.

dqwu:
@jayeshkrishna 
You can list some important PRs (with links) in E3SM-Project:master
that might need go to NCAR:master

dqwu:
@jedwards4b 
When configuring latest PIO with C++ support, on Linux workstations we use:
CC=mpicc CXX=mpicxx FC=mpifort cmake ...

On Cori we use
CC=cc CXX=CC FC=ftn cmake ...

I remember if we forget to put CXX=**** it will cause build errors.

jedwards4b:
Putting #include <mpi.h> explicitly in io_sdecomp_regex.cpp solves the build issue @jayeshkrishna thank you.
@edhartnett there is no c++ API here, only some code for parsing regular expressions written in c++, I think that the code is useful and will look into making it an optional feature.

src/clib/pio_rearrange.c
```diff
@@ -332,9 +338,16 @@ int create_mpi_datatypes(MPI_Datatype mpitype, int msgcnt,
         if (mcount[i] > 0)
         {
             int len = mcount[i] / blocksize;
-            int displace[len];
+            int *displace = NULL;
```
jedwards4b: I'm not sure this change makes any sense.  First len can never be < 1, second suppose that len were == 0 you would have a bug in the following code where displace is used and freed.

jayeshkrishna:
Replaced by PR #1410 (Since master is being updated in the fork it is easier to merge/discuss changes in a snapshot )

Issue #1364, opened 2019-03-20
Add fortran build to autotools build system


jedwards4b:
I think that being able to configure and build flib and tests with autotools is a good alternative to building cmake in travis.   We are testing the cmake build nightly so it's not ideal but it is tested.

edhartnett:
One area that still needs to be resolve is the generation of code in src/flib with the perl script genf90.pl. This is in a separate github project?

What would be easiest would be to just include a copy of that script in our build, as we do with the script that builds the fortran test code in tests/general.

jedwards4b:
genf90.pl is in a separate repository - you will see it referenced in src/flib/CMakeLists.txt 
would it work to add it as a git subtree or submodule?   The script for fortran tests is used only for
fortran tests will genf90.pl is used by a number of other projects, although there has been little change in that script in several years I don't want to diverge.

edhartnett:
Seems like we need to include a copy of that script in the build.

If we rely on another repo to supply it, then what happens with old versions of PIO? For example, let's say we do a 2.4.2 release and it uses the current genf90.pl.

Then, a year later, someone wants to build 2.4.2 again. But genf90.pl may have changed, and so they won't get the same 2.4.2 that they got a year before, which means that what we thoroughly tested is not what they are building.

So we need to package a copy of genf90.pl with each release, and it will remain fixed for that release, though future releases may update genf90.pl as and if it changes. However, each release will have a fixed version of genf90.pl which will not change.

We could accomplish this with a submodule, however that's really overkill. I will just put a copy of the script in the code directory and everything will work fine.

jedwards4b:
ok

PR #1378, opened 2019-03-26
autotools fortran build

edhartnett:
Part of #1364.

The autotools fortran build. Use the --enable-fortran option to configure to get the fortran library build.

There are still some issues to clear up but this is a good first cut. With this PR I also turn on the fortran build (along with address sanitizer) and do a make distcheck.


jedwards4b:
I don't think that all of these pregenerated files should be added.

edhartnett:
I agree, it's just a temporary step while we work out what happens with the generating perl script. See #1364 .

PR #1384, opened 2019-03-29
fixed memory leaks on error

edhartnett:


edhartnett:
Stand by I will take another stab at this...

PR #1389, opened 2019-04-04
switch cmake build to use malloc not bget by default

edhartnett:
Attempting to sort out memory issues before making other changes...


jedwards4b:
The reason these tests fail with malloc but not bget - is that they are not implemented with bget.  
See tests/cunit/CMakelists.txt line 95

edhartnett:
OK, sorry for the confusion.

But in any case, I can't reproduce these. And we can see from the travis build of this branch, the autotools and cmake builds are both run.

Also I have a jenkins matrix that runs for all combinations of 3 netcdf and 2 pnetcdf versions. And everything is passing.

Perhaps the problem is intermittent?

jedwards4b:
I can reproduce the problem consistently, but it is intermittent in that it tends to be the second comp task which is left behind, but occasionally its the first comp task.

edhartnett:
I will take this PR down since this change is in PR #1391 as well.

PR #1391, opened 2019-04-05
Msg overwrite fix

jedwards4b:
The issue was that all components were writing to the same memory location on the io root and so the value of msg could be overwritten at any point in the pio_msg_handler2.   The solution is to provide a unique location for each component to send messages to.   

All tests now pass with gnu/openmpi - test 30 test_darray_async_many is still timing out with
intel/mpt.

### DIFF
--- a/src/clib/pio_msg.c
+++ b/src/clib/pio_msg.c
@@ -2521,7 +2521,7 @@ int pio_msg_handler2(int io_rank, int component_count, iosystem_desc_t **iosys,
         {
             my_iosys = iosys[cmp];
             LOG((1, "about to call MPI_Irecv union_comm = %d", my_iosys->union_comm));
-            if ((mpierr = MPI_Irecv(messages+cmp, 1, MPI_INT, my_iosys->comproot, MPI_ANY_TAG,
+            if ((mpierr = MPI_Irecv(&(messages[cmp]), 1, MPI_INT, my_iosys->comproot, MPI_ANY_TAG,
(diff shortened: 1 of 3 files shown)


src/clib/pio_msg.c
```diff
@@ -2521,7 +2521,7 @@ int pio_msg_handler2(int io_rank, int component_count, iosystem_desc_t **iosys,
         {
             my_iosys = iosys[cmp];
             LOG((1, "about to call MPI_Irecv union_comm = %d", my_iosys->union_comm));
-            if ((mpierr = MPI_Irecv(&msg, 1, MPI_INT, my_iosys->comproot, MPI_ANY_TAG,
+            if ((mpierr = MPI_Irecv(messages+cmp, 1, MPI_INT, my_iosys->comproot, MPI_ANY_TAG,
```
edhartnett: How about the less confusing notation messages[cmp]?

edhartnett:
OK, brilliant that you figured this out! ;-)

src/clib/pio_msg.c
jedwards4b: changed

jedwards4b:
So what is the process now - do I merge to master when travis completes?

edhartnett:
What I have been doing is submitting PRs, waiting until travis succeeds, and then merging them to develop and waiting a day. See #1356 for discussion.

I believe we should change to merges after travis completes, but in deference to everyone else I am still following the rules we set up a couple of years ago, using the develop branch.

Issue #1219, opened 2017-09-22
Improve documentation


edhartnett:
Improve the introduction in the docs. Need to describe what PIO is.

edhartnett:
Well this is an oldie but a goodie.

I am putting together an extended abstract on PIO performance. I am doing some diagrams there, so perhaps they might be useful for the PIO documentation.

Here's one:
![image](https://user-images.githubusercontent.com/4533212/56972415-2c715700-6b28-11e9-8c89-450f1e045439.png)

edhartnett:
Here's another, comments welcome:

![PIO Intracomm](https://user-images.githubusercontent.com/4533212/56985994-dfe94400-6b46-11e9-9f8d-963d34d74c3e.png)

edhartnett:
Here's another:
![PIO Async](https://user-images.githubusercontent.com/4533212/56992346-0fa04800-6b57-11e9-80b7-130e8ff9275b.png)

edhartnett:
And another...

![PIO Decomposition](https://user-images.githubusercontent.com/4533212/56994415-3c0a9300-6b5c-11e9-8cdc-abc8244888ac.png)

edhartnett:
And more!

![PIO Async Multiple Computation Units](https://user-images.githubusercontent.com/4533212/57016194-0e623000-6bd6-11e9-8eb6-2595780bf591.png)

edhartnett:
Soon we are going to introduce NOAA modelers to PIO. I would like to do some documentation work first, so that they may find good answer while their interest is high.

Before I close this ticket, I would like to do the following:
* Separate documentation into a language-neutral PIO Users Guide, and a PIO C Reference and a PIO Fortran Reference. The users guide will include basically all the text documentation under doc. The C and Fortran References will contain the doxygen documentation for the functions. Of course all three documents retain cross-links to each other as appropriate.
* Eliminate all doxygen warnings.
* Incorporate the diagrams shown above into the documentation.

Of course the documentation will continue to evolve and improve, but when these changes are complete I will close this ticket and future documentation tasks will be covered in other tickets.

edhartnett:
An update:
* all doxygen warnings have been fixed for all C code and for the Fortran code that is scanned for documentation.
* Not all fortran code is scanned for documentation, only these files (from configure.ac);
[...]

PR #1404, opened 2019-05-01
more doc fixes

edhartnett:
Part of #1219


edhartnett:
I am actually going to change test_perf2 so it just uses whatever number of processors it runs on. More changes coming tomorrow... ;-)

PR #1413, opened 2019-05-12
added call to nc_finalize() to free any residual netCDF memory use

### LINKED ISSUE #1406: we need to call nc_finalize() in PIOc_finalize()
There is an nc_finalize() call that frees some memory allocated by curl and other components of the netCDF stack. It is optional, but if not called, causes false failures on memory checking runs of PIO tests.

So we just need to call it when we are all done with the netCDF library.

edhartnett:
Fixes #1406.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -140,6 +140,21 @@ AC_CHECK_LIB([netcdf], [nc_create], [], [AC_MSG_ERROR([Can't find or link to the
 # Do we have nc_finalize() in the netcdf library?
 AC_CHECK_FUNCS(nc_finalize)
 
+# But nc_finalize should only be used after 4.6.2.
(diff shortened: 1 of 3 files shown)


edhartnett:
OK a problem is that this breaks test for netcdf-4.6.2. The reason is a bug in the provenience memory cleanup code. I am pondering what to do here...

edhartnett:
OK, turns out this also breaks with netCDF 4.7.0. I am investigating...

edhartnett:
OK, nc_finalize() does not yet do what we need it to do for it to be added to PIOc_finalize().

The problem is that nc_finalize() will cause netCDF to break if you call nc_finalize() and then try to create or open a netCDF file.

For now, it should be used at the end of the test program, not in any PIO code.

PR #1414, opened 2019-05-12
bug fix: allocate my_proc_list after checking component_count

wkliao:
Test program tests/cunit/test_async_simple.c, line 70 calls PIOc_init_async with argument component_count set to -1.
https://github.com/NCAR/ParallelIO/blob/bd08dcc9ce4f30b2c03f5d705496a1630a4a92d8/tests/cunit/test_async_simple.c#L70-L71

This negative value can cause an error for variable my_proc_list.
https://github.com/NCAR/ParallelIO/blob/bd08dcc9ce4f30b2c03f5d705496a1630a4a92d8/src/clib/pioc.c#L1336


edhartnett:
Thanks! Good catch.

Issue #1424, opened 2019-05-15
shall we introduce error mode PIO_REDUCE_ERROR?


edhartnett:
Once change proposed in #1410 is the addition of a new error handling code: 

```
/** Error codes are reduced across all processes. */
￼     PIO_REDUCE_ERROR = (-53),

```

The code for this is in pioc_support.c, starting at line 534.

@jedwards4b the question is, do we add this error handler or not?

@jayeshkrishna can you give us an idea of how this is helpful?

jayeshkrishna:
This error handler would be useful in detecting errors across all MPI processes - https://github.com/E3SM-Project/ParallelIO/pull/146

jedwards4b:
These error handlers are intended for use in user code on compute tasks.  PIO already reduces errors to io task 0 before returning to the compute space so I don't think that any new functionality is added with this change.   @jayeshkrishna Is there something I'm missing?   Can you give an example of where this would need to be used in a user code instead of PIO_BCAST_ERROR?

jayeshkrishna:
This error handler should be able to capture errors in the non-root I/O procs, capture multiple error codes from different I/O procs etc. This is intended for developers debugging issues in PIO (not production).

edhartnett:
OK, I believe we have decided not to proceed with this.

In terms of debugging PIO, I find the current error handlers work as well as I could wish. When my test program crashes, I get the error from ioproc 0 and that tells me the line and source file, and that's what I need.

PR #1428, opened 2019-05-16
Make async mode writes really async

### LINKED ISSUE #1427: shall we really go async?
Our async mode is not really async.

When writing data, ultimately PIOc_put_vars_tc() is called. In PIOc_put_vars_tc, the following code appears at the bottom:
```
    /* Broadcast and check the return code. */
    if ((mpierr = MPI_Bcast(&ierr, 1, MPI_INT, ios->ioroot, ios->my_comm)))
        return check_mpi(NULL, file, mpierr, __FILE__, __LINE__);
    if (ierr)
        return check_netcdf(file, ierr, __FILE__, __LINE__);
    LOG((2, "PIOc_put_vars_tc bcast netcdf return code %d complete", ierr));
```

So what is happening is that the compute nodes actually wait for the data to be written, and then they get the return code.

But if we blow off the return code MPI_Bcast, then the compute processor can return from the function after transfering data to the IO processor. As soon as the MPI data transfer is complete, the compute processor can get a return code of 0 and continue on it's merry way, while the I/O processor continues to do the disk write.

This is easily accomplished in code. The above code block is commented out, and only the IO processor checks the return code of the disk write. If it fails, well, then we are all screwed, because the compute processor has continued to process, but aside from that, making this change should allow true async operation.

When I make this change and run on jet, I see no difference in write performance, but this is expected. The write does not happen faster, but the compute processor gets to compute while data is being written.

@jedwards4b can you ponder this change and tell me if you think it's OK? I will submit a PR this morning...

edhartnett:
Fixes #1427.

No longer make compute processors wait for return code from data writes, allowing them to continue processing while writes are happening.

### DIFF
--- a/src/clib/pio_getput_int.c
+++ b/src/clib/pio_getput_int.c
@@ -1254,13 +1254,16 @@ int PIOc_put_vars_tc(int ncid, int varid, const PIO_Offset *start, const PIO_Off
         /* Free malloced resources. */
         if (ndims && !stride_present)
            free(fake_stride);
+
+        if (ierr)
+            return check_netcdf(file, ierr, __FILE__, __LINE__);
     }
 
     /* Broadcast and check the return code. */
-    if ((mpierr = MPI_Bcast(&ierr, 1, MPI_INT, ios->ioroot, ios->my_comm)))
-        return check_mpi(NULL, file, mpierr, __FILE__, __LINE__);
-    if (ierr)
-        return check_netcdf(file, ierr, __FILE__, __LINE__);
+    /* if ((mpierr = MPI_Bcast(&ierr, 1, MPI_INT, ios->ioroot, ios->my_comm))) */
+    /*     return check_mpi(NULL, file, mpierr, __FILE__, __LINE__); */
+    /* if (ierr) */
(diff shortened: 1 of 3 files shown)


edhartnett:
I agree, but what is sane in this context?

The compute task has already moved on. What should the IO task do with the error?

jedwards4b:
I guess we have to assume errors are fatal.   We don't have a mechanism for the io tasks to signal the compute tasks but maybe we can report the error the next time compute tasks make contact with the io tasks?

edhartnett:
I was thinking that as well. Essentially that's what will happen if we do nothing. The IO tasks will error and exit. The next time the compute tasks try to contact, they will time out or error out because the IO communicator is gone.

I will write some tests for this, and also to examine the proposed new error mode (#1424).

edhartnett:
Assuming errors are fatal is fine. I doubt any model code in existence tries to do anything other than exit with a netcdf error code. I'm just happy if they bother to check the error code.

For netcdf errors, what can the model do, other than throw up its hands and exit?

gold2718:
I assume you mean read and write errors. There is lots of code that handles errors from inquiry functions.

jedwards4b:
yes, read and write errors - inquire functions are by necessity not async.

PR #1454, opened 2019-05-27
more documentation work, including some fortran code

edhartnett:
Part of #1219.

@jedwards4b I am starting to make changes to F90 files, and I want to get you to take a look before I proceed further.

Emacs is wanting me to indent comments to the same column as the code, is that a problem?

For example:

```
!>
!! @public
!! @ingroup PIO_getnumiotasks
!! @brief This returns the number of IO-tasks that PIO is using
!! @param iosystem : a defined pio system descriptor, see PIO_types
!! @param numiotasks : the number of IO-tasks
!<
   subroutine getnumiotasks(iosystem,numiotasks)

```
becomes:

```
  !>
  !! @public
  !! @ingroup PIO_getnumiotasks
  !! @brief This returns the number of IO-tasks that PIO is using
  !! @param iosystem : a defined pio system descriptor, see PIO_types
  !! @param numiotasks : the number of IO-tasks
  !<
   subroutine getnumiotasks(iosystem,numiotasks)

```

Other changes are going to be to remove the @brief and @details tags (not needed), and to add @param tags where needed, as well as potentially editing and adding to the descriptions.

### DIFF
--- a/src/flib/pio_kinds.F90
+++ b/src/flib/pio_kinds.F90
@@ -1,22 +1,11 @@
 !>
-!! @file pio_kinds.F90
-!! @brief basic data types 
+!! @file
+!! This module defines default numerical data types for all common data
+!! types like integer, character, logical, real4 and real8.
 !!
 !<
  module pio_kinds
 
-!BOP
-! !MODULE: pio_kinds
-!
-! !DESCRIPTION:
-!  This module defines default numerical data types for all common data
-!  types like integer, character, logical, real4 and real8.
-!
-! !REVISION HISTORY:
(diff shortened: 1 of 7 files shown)


src/flib/pio_kinds.F90
```diff
@@ -44,11 +33,4 @@ module pio_kinds
 !> Byte size of the MPI_OFFSET type.
    integer, parameter, public :: PIO_OFFSET_KIND=MPI_OFFSET_KIND
 
-!EOP
-!BOC
-!EOC
```
edhartnett: Should I put these back in? Or remove them everywhere (my desire if they are no longer used)?

Issue #1469, opened 2019-06-05
Add PIOc_get_vard_*() and PIOc_put_vard_*() functions


edhartnett:
As discussed in the recent I/O workshop, I am going to add some PIOc_get_vard() and PIOc_put_vard() functions, to make our API more netCDF-like.

These are intended to provide a better look and feel for using distributed arrays, and will be used in (some) new code instead of PIOc_setframe() and PIOc_write_darray()/PIOc_read_darray().

edhartnett:
These have been added and were in the 2.4.3 release.

PR #1470, opened 2019-06-05
new vard functions

edhartnett:
Part of #1469.

Still adding tests and then have to enable data conversion.


edhartnett:
I have finished adding tests and am gong to merge this PR as soon as it passes travis.

PR #1477, opened 2019-06-07
clean up of pio_types.h

### LINKED ISSUE #1436: pio_types.F90 is written to handle pnetcdf-only builds, but the C library doesn't support that anyway
The PIO C library requires netcdf to build.

But the pio_types.F90 code is written to accommodate pnetcdf-only builds; this can be simplified.

Also it see that it does not really use CDF5 format when using the classic library, but that is quite possible now.

edhartnett:
Fixes #1436.

We changed the library years ago so that netcdf-c is always required, pnetcdf is optional.

This brings pio_types.F90 up to date with that decision, removing the code that used to handle pnetcdf-only builds, which are no longer supported.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -14,4 +14,4 @@ SUBDIRS = src tests examples ${DOC} scripts
 
 EXTRA_DIST = CMakeLists.txt set_flags.am COPYRIGHT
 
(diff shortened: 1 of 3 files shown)


edhartnett:
I will take this down and try it in smaller pieces...

Issue #1341, opened 2019-03-06
a funny thing happened on the way to exascale...


edhartnett:
I'm not sure what is going on here. I am on branch ejh_benchmark, looking at program test_perf2.c in tests/cunit.

This is a straight-up copy of test_darray_3d.c, with a few changes. I am trying to scale up the size of the data arrays. In test_data_3d.c they are:

```
/* The length of our sample data along each dimension. */
#define X_DIM_LEN 4
#define Y_DIM_LEN 4
#define Z_DIM_LEN 4

```

I can successfully scale them up to this:

```
/* The length of our sample data along each dimension. */
#define X_DIM_LEN 128
#define Y_DIM_LEN 128
#define Z_DIM_LEN 64
```
But when I try this, I get a segfault:

```
#define X_DIM_LEN 256
#define Y_DIM_LEN 256
#define Z_DIM_LEN 64

```

When I use the debugger to find where the segfault happens, it is here:

```
Program received signal SIGSEGV, Segmentation fault.
0x0000555555570063 in box_rearrange_create (ios=0x5555557d1380, maplen=1048576, compmap=0x7fffed88d010, 
    gdimlen=0x7fffffffe0dc, ndims=3, iodesc=0x5555557e14a0) at pio_rearrange.c:1221
1221            dest_ioproc[i] = -1;

```

This is strange. The code looks like this:

```
int box_rearrange_create(iosystem_desc_t *ios, int maplen, const PIO_Offset *compmap,
                         const int *gdimlen, int ndims, io_desc_t *iodesc)
{
    int ret;

    /* Check inputs. */
    pioassert(ios && maplen >= 0 && compmap && gdimlen && ndims > 0 && iodesc,
              "invalid input", __FILE__, __LINE__);
    LOG((1, "box_rearrange_create maplen = %d ndims = %d ios->num_comptasks = %d "
         "ios->num_iotasks = %d", maplen, ndims, ios->num_comptasks, ios->num_iotasks));

    /* Allocate arrays needed for this function. */
    int dest_ioproc[maplen]; /* Destination IO task for each data element on compute task. */
    PIO_Offset dest_ioindex[maplen];    /* Offset into IO task array for each data element. */
    int sendcounts[ios->num_uniontasks]; /* Send counts for swapm call. */
    int sdispls[ios->num_uniontasks];    /* Send displacements for swapm. */
    int recvcounts[ios->num_uniontasks]; /* Receive counts for swapm. */
[...]

PR #1500, opened 2019-06-22
move some stack allocations to heap

edhartnett:
Part of #1341.


edhartnett:
Needs more work, I will try again...

PR #1502, opened 2019-06-24
Clean up of error handling macros

edhartnett:
Part of #1500.

I am trying to move some stack allocations to heap. This means we need slightly more sophisticated error handling in some functions, in order to clean up our memory before leaving the function on error.

The method is the same as is used in the netcdf-c library. An exit label is added and all allocations freed after that. Instead of returning an error code, the code calls macro EXIT() with the error code. This calls pio_err() and then goes to exit to free memory and return the error code.

Note some important parts of this idiom:
* The var "ret" is used for return codes.
* ret must be initialized to PIO_NOERR.
* All pointers that will receive memory allocations must be initialized to NULL.

### DIFF
--- a/src/clib/Makefile.am
+++ b/src/clib/Makefile.am
@@ -9,7 +9,7 @@ lib_LTLIBRARIES = libpio.la
 # for information regarding incrementing `-version-info`.
 libpio_la_LDFLAGS = -version-info 2:0:1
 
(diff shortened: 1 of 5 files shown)


edhartnett:
I think something is wrong with travis this morning. I will check back in a few hours and restart this PR.

PR #1580, opened 2019-08-07
update modules used in cheyenne cmake testing

jedwards4b:
Update the external software levels of cmake tests on cheyenne.

### DIFF
--- a/tests/cunit/CMakeLists.txt
+++ b/tests/cunit/CMakeLists.txt
@@ -68,6 +68,8 @@ if (NOT PIO_USE_MPISERIAL)
   target_link_libraries (test_pioc_fill pioc)
   add_executable (test_darray EXCLUDE_FROM_ALL test_darray.c test_common.c)
   target_link_libraries (test_darray pioc)
+  add_executable (test_darray_uneven EXCLUDE_FROM_ALL test_darray_uneven.c test_common.c)
+  target_link_libraries (test_darray_uneven pioc)
   add_executable (test_darray_frame EXCLUDE_FROM_ALL test_darray_frame.c test_common.c)
   target_link_libraries (test_darray_frame pioc)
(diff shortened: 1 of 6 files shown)


jedwards4b:
@edhartnett the failing test was added in a previous PR and seems to be passing, this PR should not have affected it at all - any idea why it fails here?

tests/cunit/CMakeLists.txt
```diff
@@ -233,6 +235,10 @@ else ()
     EXECUTABLE ${CMAKE_CURRENT_BINARY_DIR}/test_darray
     NUMPROCS ${AT_LEAST_FOUR_TASKS}
     TIMEOUT ${DEFAULT_TEST_TIMEOUT})
+  add_mpi_test(test_darray_uneven
```
edhartnett: Well here is the issue. You added this test to the cmake build and it is failing.

tests/cunit/CMakeLists.txt
```diff
@@ -68,6 +68,8 @@ if (NOT PIO_USE_MPISERIAL)
   target_link_libraries (test_pioc_fill pioc)
   add_executable (test_darray EXCLUDE_FROM_ALL test_darray.c test_common.c)
   target_link_libraries (test_darray pioc)
+  add_executable (test_darray_uneven EXCLUDE_FROM_ALL test_darray_uneven.c test_common.c)
```
edhartnett: This test does not exist in the code. If you created it, try git add test_darray_uneven.c. If not, then delete all references to it in cmake.

jedwards4b: oops - I thought that I had added it. I will correct - thanks.

tests/cunit/CMakeLists.txt
jedwards4b: It works locally - how do I see details of the travis run?

PR #1581, opened 2019-08-09
must allocate on ioroot for serial io

jedwards4b:
When writing with netcdf or netcdf4c the ioroot buffer must be maxiobuflen so that it can receive from all iotasks.

### DIFF
--- a/src/clib/pio_darray.c
+++ b/src/clib/pio_darray.c
@@ -265,7 +265,7 @@ PIOc_write_darray_multi(int ncid, const int *varids, int ioid, int nvars,
      * large as the largest used to accommodate this serial io
      * method.  */
     rlen = 0;
-    if (iodesc->llen > 0)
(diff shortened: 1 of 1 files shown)


src/clib/pio_darray.c
```diff
@@ -265,7 +265,7 @@ PIOc_write_darray_multi(int ncid, const int *varids, int ioid, int nvars,
      * large as the largest used to accommodate this serial io
      * method.  */
     rlen = 0;
-    if (iodesc->llen > 0)
+    if (iodesc->llen > 0 || (file->iotype==PIO_IOTYPE_NETCDF || file->iotype == PIO_IOTYPE_NETCDF4C) && ios->iomaster)
```
jedwards4b: I'm not sure how this code was lost - it's documented in the comment above.

PR #1609, opened 2019-11-12
move reliance on ifdefs out of pio.h

### LINKED ISSUE #1608: cannot ifdef on _NETCDF, _NETCDF4, or _PNETCDF in pio.h
We cannot depend on symbols like _NETCDF, _NETCDF4, or _PNETCDF to be defined by applications using PIO. So we cannot depend on them, as we do, in pio.h:

```
#ifdef _NETCDF
#include <netcdf.h>
#ifdef _NETCDF4
#include <netcdf_par.h>
#endif
#endif
#ifdef _PNETCDF
#include <pnetcdf.h>
#endif

#ifndef MPI_OFFSET
/** MPI_OFFSET is an integer type of size sufficient to represent the
 * size (in bytes) of the largest file supported by MPI. In some MPI
 * implementations MPI_OFFSET is not properly defined.  */
#define MPI_OFFSET  MPI_LONG_LONG
#endif
```

pio.h is our public header, and does not (and can not) include config.h. So it would be up to the user of PIO to know they have to turn on these macros. That's not good.

edhartnett:
Fixes #1608 

In this PR I move some code that depends on ifdefs out of pio.h.

pio.h is our public header, we cannot count on these ifdefs being defined by the user. We must confine out use of these ifdefs to files that include config.h, and our public header cannot include that file.

### DIFF
--- a/src/clib/pio.h
+++ b/src/clib/pio.h
@@ -18,13 +18,6 @@
 
 #include <netcdf.h>
 
-#ifndef MPI_OFFSET
-/** MPI_OFFSET is an integer type of size sufficient to represent the
- * size (in bytes) of the largest file supported by MPI. In some MPI
- * implementations MPI_OFFSET is not properly defined.  */
-#define MPI_OFFSET  MPI_LONG_LONG
-#endif
-
 /** PIO_OFFSET is an integer type of size sufficient to represent the
  * size (in bytes) of the largest file supported by MPI. This is not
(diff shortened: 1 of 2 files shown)


edhartnett:
Any additional NC_ defines needed will be found when pio_internal.h is included.

Since they are not used in pio.h, they don't need to be found by pio.h.

PR #1616, opened 2019-11-29
move structs from pio.h to pio_internal.h

### LINKED ISSUE #1612: pio.h exposes implementation details
`pio.h` contains several internal structures (`var_desc_t`, `io_region`, `io_desc_t`, `iosystem_desc_t`, `wmulti_buffer`, `file_desc_t`, ...) that could be moved to `pio_internal.h`.

This would eliminate the need to install `uthash.h` (see #1551). Installing `uthash.h` is problematic because if a user has `uthash.h` installed already the user's copy will be overwritten.

I can create a pull request with these changes if you think it's a good idea.

edwardhartnett:
Fixes #1612 

Move some structs to pio_internal.h from pio.h, which allows us to keep uthash.h internal to PIO. It will no longer be installed with PIO.


edwardhartnett:
I'm going to try again...

PR #1618, opened 2019-12-03
For netCDF integration opens/creates of PIO_IOTYPE_NETCDF4C, don't automatically turn on deflation for every var

### LINKED ISSUE #1617: Allow netcdf-4/HDF5 serial access without compression
THe huge performance problems in PIO_IOTYPE_NETCDF4C files is caused by compression. We need to let users create files without compression!

Now this is the case with files created through netCDF integration. If a file is opened like this:

`            if (nc_create(FILE_NAME, NC_PIO|NC_NETCDF4, &ncid)) PERR;`

The a sequential netCDF-4/HDF5 file will be created, but nc_def_var_deflate() will not be automatically called for each var. Instead, the user muse explicitly call nc_def_var_deflate() for each var that they want compressed. This matches normal netCDF API use so will be more familiar to netCDF users.

I will have a PR for this momentarily.

edwardhartnett:
Fixes #1617 

After this PR, when opening a netCDF-4/HDF5 file in sequential mode using netCDF integration, then deflate will no longer be automatically turned on.

Existing PIOc_* code is not affected by this change. Only new code, written to use the netCDF integration features, will be affected.

Also in this PR I added a performance testing program tests/ncint/tst_ncint_perf.c.

### DIFF
--- a/src/clib/pio.h
+++ b/src/clib/pio.h
@@ -584,6 +584,11 @@ typedef struct file_desc_t
     /** True if this task should participate in IO (only true for one
      * task with netcdf serial files. */
     int do_io;
+
+    /** True if this file was opened with the netCDF integration
+     * feature. One consequence is that PIO_IOTYPE_NETCDF4C files will
+     * not have deflate automatically turned on for each var. */
+    int ncint_file;
 } file_desc_t;
 
 /**
--- a/src/clib/pio_nc.c
+++ b/src/clib/pio_nc.c
@@ -2278,8 +2278,10 @@ PIOc_def_var(int ncid, const char *name, nc_type xtype, int ndims,
         PLOG((3, "defined var ierr %d file->iotype %d", ierr, file->iotype));
 
 #ifdef _NETCDF4
-        /* For netCDF-4 serial files, turn on compression for this variable. */
-        if (!ierr && file->iotype == PIO_IOTYPE_NETCDF4C)
+        /* For netCDF-4 serial files, turn on compression for this
+         * variable, unless this file was opened through the netCDF
+         * integration feature. */
+        if (!ierr && file->iotype == PIO_IOTYPE_NETCDF4C && !file->ncint_file)
             ierr = nc_def_var_deflate(file->fh, varid, 0, 1, 1);
 
         /* For netCDF-4 parallel files, set parallel access to collective. */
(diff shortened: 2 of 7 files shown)


jedwards4b:
Doesn't this conflict with the meaning of PIO_IOTYPE_NETCDF4C - the C stands for compression, the user expects compression to be enabled.   If you are looking to enable netcdf4 with serial IO then maybe the user should use PIO_IOTYPE_NETCDF and use the mode flag to set netcdf4 format.

jedwards4b:
I don't understand why you merged this.

edwardhartnett:
Sorry, I didn't see your comment.

This has no effect at all on the PIOc_* calls for PIO. In other works, no existing code will be affected.

What this means is that, when using the netCDF integration (i.e. using PIO with normal netcdf-c calls), the library will not automatically apply compression. (The user never sees or uses PIO_IOTYPE_NETCDF4C, so that doesn't matter.)

Here's how it works:

```
        /* Initialize the intracomm. The IO task will not return from
         * this call until the PIOc_finalize() is called by the
         * compute tasks. */
        if (nc_def_async(MPI_COMM_WORLD, num_io_procs, NULL, COMPONENT_COUNT,
                         num_procs2, NULL, NULL, NULL, PIO_REARR_BOX, &iosysid))
            PERR;

        if (my_rank)
        {
 
            /* Create a file with a 3D record var. */
            if (nc_create(FILE_NAME, NC_PIO|NC_NETCDF4, &ncid)) PERR;
            if (nc_def_dim(ncid, DIM_NAME_UNLIMITED, dimlen[0], &dimid[0])) PERR;
            if (nc_def_dim(ncid, DIM_NAME_X, dimlen[1], &dimid[1])) PERR;
            if (nc_def_dim(ncid, DIM_NAME_Y, dimlen[2], &dimid[2])) PERR;
            if (nc_def_var(ncid, VAR_NAME, NC_INT, NDIM3, dimid, &varid)) PERR;
            if (nc_enddef(ncid)) PERR;
...


```

In the code above, the nc_def_async() call initialized PIO and netCDF, and sets up the async iosystem.

Then the netCDF calls create and define a simple file, as usual. Note the NC_PIO in the nc_create() call.

Before my recent PR, this code would have turned on compression in the var in the file. But clearly this is not correct, because in the netCDF API you don't get compression without explicitly asking for it. So for netCDF integration to work, it cannot assume that all netCDF-4/HDF5 files want compression.

However, we can still get compression by explicitly calling nc_def_var_deflate() for the var.

jedwards4b:
okay - thanks for the explination

PR #1653, opened 2020-06-08
Short type in fortran

jedwards4b:
Adds support for a PIO_short datatype in the fortran interface.


jedwards4b:
@edwardhartnett Can you have a look at the travis CI results - I'm not sure how to address the failure. 

I've tested with intel/19.0.5 and pgi/20.5 and openmpi/4.0.3 on cheyenne - all tests pass in both cases.

jedwards4b:
all tests pass with gnu/9.1.0 as well.

jedwards4b:
@edwardhartnett I need help with travis - I tried the bionic distro but that results in not being able to find netcdf.   And the Xenial doesn't seem to find netcdf.inc.

PR #1656, opened 2020-07-13
Ejh doc fixes

edwardhartnett:
Documentation fixes to get warning-free documentation build.

### DIFF
--- a/src/flib/pio_types.F90
+++ b/src/flib/pio_types.F90
@@ -153,8 +153,8 @@ module pio_types
   double precision, public, parameter :: PIO_FILL_DOUBLE = nf_fill_double; !< double fill value
 
   enum, bind(c)
-     enumerator :: PIO_rearr_comm_p2p = 0 !< rearranger peer to peer
-     enumerator :: PIO_rearr_comm_coll    !< rearranger collective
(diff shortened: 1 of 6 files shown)


src/flib/pio_types.F90
```diff
@@ -153,8 +153,8 @@ module pio_types
   double precision, public, parameter :: PIO_FILL_DOUBLE = nf_fill_double; !< double fill value
 
   enum, bind(c)
-     enumerator :: PIO_rearr_comm_p2p = 0
-     enumerator :: PIO_rearr_comm_coll
+     enumerator :: PIO_rearr_comm_p2p = 0 !< rearranger peer to peer
+     enumerator :: PIO_rearr_comm_coll    !< rearranger collective
```
edwardhartnett, line 157: @jedwards4b what are these really? I couldn't figure it out so just guessed...

jedwards4b: Options for the rearr_opts.comm_type variable: PIO_rearr_comm_coll means to use the MPI_ALLTOALLW function of the mpi library and PIO_REARR_COMM_P2P means to do point-to-point communications using mpi send and recv calls.   In general P2P is faster, but your mileage may vary.

edwardhartnett: Thanks! I have updated the code with those comments. New release will be out shortly...

PR #1662, opened 2020-07-16
Fix multiple definition errors with GCC>=10.1.0 

### LINKED ISSUE #1651: gcc 10.1.0 compile error and solution
A change in default flags in the gcc compiler results in errors when building pio.  
The error is: pio_error.h:83: multiple definition of `resultlen'; .libs/libpioc_la-bget.o:/glade/scratch/sghosh/gnu10/pio/72539/pio-2.4.4/src/clib/./pio_error.h:83: first defined here

The solution is to add the flag -fcommon

markjolah:
This patch fixes #1651 by properly separating declarations of global variables in C headers from their definitions which are now moved to a separate .c compilation unit.  The original code cause the C compiler to add a new (strong) symbol for `err_buffer` and `resultlen` to every compilation unit that includes `pio_error.h`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -243,6 +243,7 @@ add_subdirectory (src)
 
 # Custom "piotests" target (builds the test executables)
 add_custom_target (tests)
+add_dependencies (tests pioc piof)
 
 # Custom "check" target that depends upon "tests"
 add_custom_target (check COMMAND ${CMAKE_CTEST_COMMAND})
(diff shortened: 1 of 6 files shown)


markjolah:
Hmm... not sure why those test programs don't find the symbols from `pioc` target.  I ran into some issues with parallel builds because the CMake didn't establish build dependencies between the tests and the libraries.  I will push those changes to see if they can fix the CI build error.  Everything was working for me with GCC-10.1.0 now, but maybe the CI isn't happy still?

edwardhartnett:
The reason this is failing is that you have fixed the CMake build, but not the autotools build. Both builds are checked by Travis CI.

So you need to edit Makefile.am in the same directory as pio_error.c, and add it to the list of files there as well.

edwardhartnett:
Well this is a good fix but is missing the Makefile.am part, so can't be merged. I will put this on a different branch that I can edit, and add the Makefile fix...

PR #1667, opened 2020-07-22
Now allow deflate to be used with IOTYPE NETCDF4P

### LINKED ISSUE #1642: Starting with netcdf-c-4.7.4 zlib compression may be used with NETCDF4P IOTYPE
We have some fortran tests that need to be changed. From ncdf_tests.F90:

```
    print*, 'testing PIO_def_var_deflate' 
    shuffle = 0
    deflate = 1
    deflate_level = 2
    deflate_level_2 = 4
    ret_val = PIO_def_var_deflate(pio_file, pio_var, shuffle, deflate, &
         deflate_level)

    ! Should not have worked except for netCDF-4/HDF5 serial.
    if (iotype .eq. PIO_iotype_netcdf4c .and. ret_val .ne. PIO_NOERR) then
       err_msg = "Could not turn on compression for variable foo2222"
       call PIO_closefile(pio_file)
       return
    else if (iotype .eq. PIO_iotype_pnetcdf .and. ret_val .eq. PIO_NOERR) then
       err_msg = "Did not get expected error when trying to turn deflate on for pnetcdf file"
       call PIO_closefile(pio_file)
       return
    else if (iotype .eq. PIO_iotype_netcdf .and. ret_val .eq. PIO_NOERR) then
       err_msg = "Did not get expected error when trying to turn deflate on for netcdf classic file"
       call PIO_closefile(pio_file)
       return
    else if (iotype .eq. PIO_iotype_netcdf4p .and. ret_val .eq. PIO_NOERR) then
       err_msg = "Did not get expected error when trying to turn deflate on for parallel netcdf-4 file"
       call PIO_closefile(pio_file)
       return
    end if

```

edwardhartnett:
Starting with netcdf-c-4.7.4 parallel writes can also use compression, a feature NOAA has found very, very helpful.

In this PR I take away the code in PIO that prevented this setting, and add a test which demonstrates that PIOc_def_var_deflate() can now be used with IOTYPE NETCDF4P.

Fixes #1642


edwardhartnett:
Yes, we require 4.7.4. Not because of this change, but because of changes in the behavior of some netcdf functions, which broke some of our tests.

So using older versions of netcdf-c will cause tests to break.

edwardhartnett:
@jedwards4b I was incorrect. netcdf-c-4.7.4 is needed *only* if netcdf integration is used (with --enable-netcdf-integration).

Without netCDF integration, previous versions work.

Recently I lost my CI system in a tragic boating accident. I'm recreating it now.

jedwards4b:
Thanks for that clarification.  It looks like you are replacing Travis CI with git actions?

edwardhartnett:
I'm experimenting with GitHub actions. I'll leave travis there as well, for now. ;-)

edwardhartnett:
I am going to take this down while I get testing with different versions of netCDF working.

Issue #1670, opened 2020-07-23
Issue when appending existing files


jedwards4b:
I am running into an error when appending to an existing netcdf file and have attached a simple example below.
When the file is initially created and written all works fine, but then the file is closed and reopened to append.
The issue is that in the vardesc data structure the rec_var variable is not initialized, this causes the 1d and 2d variables to 
be put into the same wmb data structure which leads to an error at pio_darray.c:192

`/*piotest uses mpi but runs on 1 task*/
#include <stdio.h>
#include <mpi.h>
#include <pio.h>

main(int argc, char **argv)
{
    int node;
    int iosys;
    int ncid;
    int iotype = PIO_IOTYPE_NETCDF;
    int dimid[2], varid[2];
    int i;
    int dimsize=42;
    int var[dimsize];
    PIO_Offset compmap[dimsize];
    int ioid;
    int ierr;

    MPI_Init(&argc,&argv);
    MPI_Comm_rank(MPI_COMM_WORLD, &node);
    if (node > 0){
	printf("Test intended for a single task\n");
	MPI_Abort(MPI_COMM_WORLD, -1);
    }
    ierr = PIOc_Init_Intracomm( MPI_COMM_WORLD, 1, 1, 0, PIO_REARR_SUBSET, &iosys);
    if(ierr) printf("%d %d\n",__LINE__,ierr);
    ierr = PIOc_createfile(iosys, &ncid, &iotype, "test1.nc", PIO_CLOBBER);
    if(ierr) printf("%d %d\n",__LINE__,ierr);
    ierr = PIOc_def_dim(ncid, "dim1", PIO_UNLIMITED, dimid);
    if(ierr) printf("%d %d\n",__LINE__,ierr);
    ierr = PIOc_def_dim(ncid, "dim2", (PIO_Offset) dimsize, dimid+1);
    if(ierr) printf("%d %d\n",__LINE__,ierr);
    ierr = PIOc_def_var(ncid, "var1", PIO_INT, 1, dimid+1, varid);
    if(ierr) printf("%d %d\n",__LINE__,ierr);
    ierr = PIOc_def_var(ncid, "var2", PIO_INT, 2, dimid, varid+1);
    if(ierr) printf("%d %d\n",__LINE__,ierr);
    ierr = PIOc_enddef(ncid);
    if(ierr) printf("%d %d\n",__LINE__,ierr);

    for(i=0;i<dimsize;i++){
	var[i] = i;
	compmap[i] = i+1;
[...]

PR #1671, opened 2020-07-24
fix initialization of var rec_var

jedwards4b:
Fixes issue #1670  logic for initializing rec_var was causing multi-dimensional variable initialization to be overwritten.


jedwards4b:
@edhartnett are these tests failing because you are actively working on them?  All my tests on cheyenne pass. 
Also I have some improvements to the cmake information in libpio.settings - would you like another PR or can I add them to this one?

jedwards4b:
@edwardhartnett

edwardhartnett:
The travis one is the only one you need to care about. But in it many tests in the cmake build are failing...

edwardhartnett:
@jedwards4b do the improvement to the build summary in a separate PR. Let's try and get this one working...

edwardhartnett:
The failures may be from the address sanitizer. That is, there may be a memory leak with your changes.

jedwards4b:
Looking at the test results it looks like the address sanitizer is also used in the autoconf build where all tests are passing.

jedwards4b:
If I build with gnu and openmpi and use the  -fsanitize=address -fno-omit-frame-pointer flags it also fails on my local system, but it also fails with the current master and nothing in the output indicates a problem with my change
103: Indirect leak of 4 byte(s) in 1 object(s) allocated from:
103:     #0 0x2adf6ae16e58 in __interceptor_malloc ../../../../libsanitizer/asan/asan_malloc_linux.cc:144
103:     #1 0x2adf726e7018  (/glade/u/apps/ch/opt/gnu/9.1.0/lib64/libasan.so.5.0.0+0x248018)
103:     #2 0x616000000000  (<unknown module>)

edwardhartnett:
And yet it worked for me...

Weird. I will take a look at this on Monday morning...

edwardhartnett:
@jedwards4b I cleaned up the CI and merged master. Let's see what happens.

src/clib/pioc_support.c
```diff
@@ -2422,7 +2424,9 @@ inq_file_metadata(file_desc_t *file, int ncid, int iotype, int *nvars,
                     (*rec_var)[v] = 0;
 
             }
+	    PLOG((2,"varid %d var_ndims %d nunlimdims %d rec_var %d unlimdimids[0]=%d var_dimids[0]=%d\n",v,var_ndims, nunlimdims, (*rec_var)[v], unlimdimids[0], var_dimids[0]));
```
edwardhartnett, line 2427: OK, could this PLOG line of code be causing your problems? In it you reference some [0] array locations, perhaps they are not always there and that is what the memory checker is complaining about?

edwardhartnett:
This needs a test as well.

tests/cunit/CMakeLists.txt
```diff
@@ -236,6 +239,10 @@ else ()
     EXECUTABLE ${CMAKE_CURRENT_BINARY_DIR}/test_darray
     NUMPROCS ${AT_LEAST_FOUR_TASKS}
     TIMEOUT ${DEFAULT_TEST_TIMEOUT})
+  add_mpi_test(test_darray_append
```
edwardhartnett, line 242: Can you add this to the Makefile.am as well?

edwardhartnett:
@jedwards4b this looks good to go. Do you want me to merge it?

PR #1672, opened 2020-07-24
applied Jim's changes to test CI

edwardhartnett:


edwardhartnett:
@jedwards4b this seems to be the change that you want, and it's passing CI with the current master.

Can we get a test for this change?

edwardhartnett:
I will take this down since you got it in your branch and PR.

PR #1691, opened 2020-07-31
turning on netcdf integration in github testing, also now testing against specific versions of netcdf-c and pnetcdf

edwardhartnett:
More testing!


edwardhartnett:
This work was merged in #1733

PR #1692, opened 2020-08-03
1642 netcdf4p deflate

nichannah:


src/flib/pio_nf.F90
```diff
@@ -1709,7 +1709,7 @@ integer function def_var_chunking(file, vardesc, storage, chunksizes) result(ier
     type (var_desc_t), intent(in) :: vardesc
     integer, intent(in) :: storage
     integer, intent(in) :: chunksizes(:)
-    integer(C_INT) :: cchunksizes(PIO_MAX_VAR_DIMS)
+    integer(C_SIZE_T) :: cchunksizes(PIO_MAX_VAR_DIMS)
```
edwardhartnett, line 1712: Nice catch! Thanks!

src/clib/pio_nc4.c
```diff
@@ -84,11 +84,8 @@ PIOc_def_var_deflate(int ncid, int varid, int shuffle, int deflate,
     if (ios->ioproc)
     {
 #ifdef _NETCDF4
-        if (file->iotype == PIO_IOTYPE_NETCDF4P)
-            ierr = NC_EINVAL;
-        else
-            if (file->do_io)
-                ierr = nc_def_var_deflate(file->fh, varid, shuffle, deflate, deflate_level);
+        if (file->do_io)
+            ierr = nc_def_var_deflate(file->fh, varid, shuffle, deflate, deflate_level);
```
edwardhartnett, line 88: I actually already have this fix, and a test for it, standing by to merge. I'm just getting the testing situation a bit more sorted out with GitHub actions, then I will merge the fix and test.

nichannah: OK - would you like me to pull the latest and remove this commit?

edwardhartnett: No we can merge this if it passes all testing, then I will merge my test for it.

src/flib/pio_nf.F90
edwardhartnett: Actually, before I merge this, can you tell me about it?

Also, is there a test where this change is demonstrated?

nichannah: The C function takes type PIO_Offset * for the chunksizes which elsewhere in the code corresponds to a C_SIZE_T array type in the Fortran interfaces. Which makes sense - chunksize are array sizes (hence size_t).  

My reading of it was that this code had been copied from elsewhere (e.g. PIO_def_var) where the array argument (e.g. dimids) holds indices rather than sizes. This also accounts for the change on line 1727, the -1 is not needed since we are dealing with sizes.

edwardhartnett: OK, I added a test and applied the suggested changes, and had to make a few more changes as well. ;-)

It was all merged as part of #1738

src/clib/pio_nc4.c
edwardhartnett: This change got merged some time ago.

edwardhartnett:
@nichannah thank you for your contribution! These changes have all been merged to master, so I will close this PR. I see by the discussion in #1642 that you are using the compression and it is working, which is great. I will close this PR.

PR #1700, opened 2020-08-06
disallow netcdf-c-4.7.2 for both cmake and autotools builds

### LINKED ISSUE #1694: PIO does not build with netcdf-c-4.7.2 due to lack of ncexternl.h file in netcdf distribution
NetCDF 4.7.2 has a bug in the parallel I/O build, in that netcdf_par.h includes ncexternl.h, but this header file is not installed.

One way to get around this would be for PIO to have its own copy of ncexternl.h from netcdf-c.

edwardhartnett:
Fixes #1694


edwardhartnett:
I would like to tighten up the fortran testing, but I need to solve #1704 first. I will merge this code. The zlib compression with IOTYPE NETCDF4P is well tested in C.

Issue #1654, opened 2020-07-10
need some netCDF integration functions to support using deflate


edwardhartnett:
The following functions need netCDF integration functions:

    nc_def_var_deflate,
    nc_def_var_fletcher32,
    nc_def_var_chunking,
    nc_def_var_endian,

edwardhartnett:
These are complete, but the nc_inq_var_chunking() function cannot work until we get a new version of the netcdf-c dispatch table. See https://github.com/Unidata/netcdf-c/issues/1836

Issue #1629, opened 2020-01-22
support new forms of compression in netCDF/HDF5 files...


edwardhartnett:
Previously only sequential zlib was available.

Now we also have parallel zlib and sequential and parallel szip.

Also, the CCR project adds bzip2 and lz4.

So there are some changes need to accommodate all these choices in PIO.

jedwards4b:
Sounds great!

edwardhartnett:
OK, we will be release CCR-1.1.0 in the next few weeks. The LZ4 filter did not work, so has been removed. However, we have the new Zstandard filter, which provides compression like zlib, but an order of magnitude faster on writes, and three times faster on reads.

We also have the bitgroom filter, which allows lossy compression by blanking out unneeded bits in the floating point or doubles.

Note that using these filters mean all readers of the data must have the filters installed.

PR #1749, opened 2020-09-09
Adding filter support to PIO

edwardhartnett:
Part of #1654 
Part of #1629 

Adding support for netCDF filters. This is needed to support all the new forms of compression, as well as to continue to support zlib compression.


edwardhartnett:
Full support for filters requires changes in the netCDF dispatch table, which means that will have to wait for a new release of netcdf-c. ;-( So I will close this PR and make the changes needed to netcdf-c, then circle back around to PIO to get filters working.

PR #1778, opened 2020-11-20
add PIOc_init_async_from_comms and test

jedwards4b:
Add another interface for PIO_init_async:
int                                                                                                                                                                                          
PIOc_init_async_from_comms(MPI_Comm world, int component_count, MPI_Comm *comp_comm, MPI_Comm io_comm, int rearranger, int *iosysidp)   

this allows the user to call pio_init with the MPI_Comm object already defined for each component and the IO tasks.

### DIFF
--- a/src/clib/pioc.c
+++ b/src/clib/pioc.c
@@ -2092,7 +2092,7 @@ PIOc_init_async_from_comms(MPI_Comm world, int component_count, MPI_Comm *comp_c
         if(comp_comm[cmp] != MPI_COMM_NULL)
             if ((ret = MPI_Comm_size(comp_comm[cmp], &(num_procs_per_comp[cmp]))))
                 return check_mpi(NULL, NULL, ret, __FILE__, __LINE__);
-        if ((ret = MPI_Allreduce(MPI_IN_PLACE, &(num_procs_per_comp[cmp]), component_count, MPI_INT, MPI_MAX, world)))
+        if ((ret = MPI_Allreduce(MPI_IN_PLACE, &(num_procs_per_comp[cmp]), 1, MPI_INT, MPI_MAX, world)))
             return check_mpi(NULL, NULL, ret, __FILE__, __LINE__);
 
     }
--- a/src/clib/pioc.c
+++ b/src/clib/pioc.c
@@ -1211,6 +1211,76 @@ PIOc_init_async_from_F90(int f90_world_comm,
 }
 
 /**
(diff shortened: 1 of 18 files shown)


edwardhartnett (approved):

edwardhartnett (approved):

tests/general/Makefile.am
```diff
@@ -67,7 +67,7 @@ TESTS = run_tests.sh
 endif # RUN_TESTS
 
 %.F90: %.F90.in
-	util/pio_tf_f90gen.pl --annotate-source --out=$@ $<
+        util/pio_tf_f90gen.pl --annotate-source --out=$@ $<
```
edwardhartnett, line 70: This needs to be a tab char, not spaces.

jedwards4b: thanks

jedwards4b:
@edhartnett Do you see what the problem is?  All tests are passing locally.

jedwards4b:
Why was .F90.in in the .gitignore file?

edwardhartnett:
Don't know, it should not have been.

src/clib/pioc.c
```diff
@@ -2085,14 +2085,14 @@ PIOc_init_async_from_comms(MPI_Comm world, int component_count, MPI_Comm *comp_c
         return check_mpi(NULL, NULL, ret, __FILE__, __LINE__);
 
     /* Get num_procs_per_comp for each comp and share with world */
-    num_procs_per_comp = (int *) calloc(component_count, sizeof(int));
+    num_procs_per_comp = (int *) malloc(component_count * sizeof(int));
```
edwardhartnett, line 2088: You are changing these so they are no longer pre-filled with zeros. Is that intentional?

jedwards4b: I am 0 filling them below - calloc seemed to be giving an error in address sanitizer.   Currently the test
pio__iosystem_async_tests passes with mpich and mpt but fails with openmpi.

edwardhartnett: OK, you want to mention that in the comments somewhere...

jedwards4b: I'm not really sure that was the problem.  Now that I have it working I could try putting back the callocs.

edwardhartnett: OK, if you want to just leave it, it's not a big deal...

Generally if I see mallocs + setting to zero I will automatically refactor to callocs whenever I tough the code anyway. ;-)

PR #1790, opened 2020-12-30
add nf_pio, netcdf now defines cdf5 format

### LINKED ISSUE #1787: Sort out where NF_PIO constant should be, and put it there
In tests/fncint/ftst_pio.f90 we have:

 ```
 ! Create a file.
  ierr = nf_create(FILE_NAME, 64, ncid)
  if (ierr .ne. nf_noerr) call handle_err(ierr)

```

The 64 is the constant that should be defined somewhere as NF_PIO.

### LINKED ISSUE #1788: netcdf now defines NF_64BIT_DATA
No need to leave this only for pnetcdf build.

jedwards4b:
Fixes #1787 
Fixes #1788

### DIFF
--- a/.github/workflows/strict_autotools.yml
+++ b/.github/workflows/strict_autotools.yml
@@ -16,6 +16,7 @@ jobs:
       FC: mpifort
       CPPFLAGS: "-I/usr/include -I/usr/local/include -I/home/runner/pnetcdf/include"
       LDFLAGS: "-L/home/runner/pnetcdf/lib"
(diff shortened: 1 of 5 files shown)


jedwards4b:
@edhartnett the change to not require pnetcdf for the fortran build requires netcdf-fortran.  I can build it for the github tests but the travis test already has a netcdf-fortran build - I guess version 4.1.3 is before NF_64BIT_DATA was added.  

I just hard coded the values copied from netcdf.inc - acceptable solution?

PR #1799, opened 2021-01-02
stole Jim's PR with subset rearranger testing and changes

### LINKED ISSUE #1798: Need an error if rearranger in pio_init_decomp inconsistant with pio_init_async
Currently just dumps core.

edwardhartnett:
Fixes #1798 

More work on subset rearranger from Jim.


edwardhartnett:
The test program test_async_perf is failing the cmake build with address sanitizer:
`AddressSanitizer failed to allocate 0x2e802000 (780148736) bytes of LargeMmapAllocator: Cannot allocate memory`

I think this program uses too much memory to be run as an ordinary test. Instead, we will build it but not run it. It's intended to be run for performance numbers, not as a normal test. For example the autotools build also builds but does not run this test as part of normal testing.

I stole your PR, and turned off this test in CMake, this should allow the PR to pass CI and get merged. Then I will run these on hera today and get some numbers. ;-)

jedwards4b:
okay, no worries.

Issue #1808, opened 2021-01-05
why is netcdf-4 parallel read so slow at low I/O processor counts?


edwardhartnett:
Look at the read time for parallel4. What's going on there?

```
tasks, nio,    rearr,  fill,   IOTYPE, write time(s),  read time(s),   data size(MB),  write(MB/s),    read(MB/s)
1000,   1,      box,    nofill, pnetcdf,           1.371,          0.273,          335.0,        244.333        1228.537
1000,   1,      box,    nofill, classic,           1.970,          0.301,          335.0,        170.070        1112.310
1000,   1,      box,    nofill, serial4,           1.443,          8.953,          335.0,        232.167          37.418
1000,   1,      box,    nofill, parallel4,         2.842,        646.826,          335.0,        117.873           0.518
1000,   1,      subset, nofill, pnetcdf,           6.887,          0.280,          335.0,         48.643        1196.262
1000,   1,      subset, nofill, classic,           1.273,          0.248,          335.0,        263.187        1352.797
1000,   1,      subset, nofill, serial4,           1.066,          8.719,          335.0,        314.377          38.424
1000,   1,      subset, nofill, parallel4,         2.229,        647.697,          335.0,        150.294           0.517
1000,   4,      box,    nofill, pnetcdf,           0.765,          0.633,          335.0,        437.924         529.572
1000,   4,      box,    nofill, classic,           1.215,          0.233,          335.0,        275.675        1437.577
1000,   4,      box,    nofill, serial4,           1.170,         11.003,          335.0,        286.420          30.445
1000,   4,      box,    nofill, parallel4,         4.732,        217.331,          335.0,         70.793           1.541
1000,   4,      subset, nofill, pnetcdf,           0.846,          0.732,          335.0,        395.913         457.737
1000,   4,      subset, nofill, classic,           2.050,          0.192,          335.0,        163.399        1741.038
1000,   4,      subset, nofill, serial4,           2.698,          8.047,          335.0,        124.161          41.633
1000,   4,      subset, nofill, parallel4,         6.454,        161.965,          335.0,         51.909           2.068
1000,   16,     box,    nofill, pnetcdf,           0.400,          0.671,          335.0,        837.720         499.206
1000,   16,     box,    nofill, classic,           1.362,          0.183,          335.0,        246.051        1832.774
1000,   16,     box,    nofill, serial4,           2.180,          8.567,          335.0,        153.671          39.103
[...]

PR #1809, opened 2021-01-07
modification to perf program output, fix perf timing, add chunksizes for netCDF-4/HDF5 files

edwardhartnett:
More work on performance programs.

Part of #1803
Part of #1808

### DIFF
--- a/tests/cunit/test_perf2.c
+++ b/tests/cunit/test_perf2.c
@@ -315,7 +315,7 @@ test_darray(int iosysid, int ioid, int num_flavors, int *flavor,
         } /* re-reading file */
         
         if (!my_rank)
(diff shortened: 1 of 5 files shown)


tests/cunit/test_perf2.c
```diff
@@ -539,8 +539,8 @@ main(int argc, char **argv)
 
     for (i = 0; i < num_io_tests; i++)
     {
-        for (r = 0; r < NUM_REARRANGERS_TO_TEST; r++)
-        /* for (r = 0; r < 1; r++) */
+        /* for (r = 0; r < NUM_REARRANGERS_TO_TEST; r++) */
```
jedwards4b, line 542: You haven't found a solution here?  It would really help if you could reproduce on a platform I have access to.

edwardhartnett: Working on it... ;-)

Issue #1728, opened 2020-08-19
cmake build on github needs to build fortran too


edwardhartnett:
Seems like the fortran build on the cmake github workflow is not being built, but it needs to be.

jedwards4b:
I think that I disabled because there is no netcdf-fortran library, but now that I think about it this isn't a requirement.
I'll have another PR soon to address this.

edwardhartnett:
IIRC netcdf.inc is used for some type constants. Which actually never change anyway, so we could just have a copy.

edwardhartnett:
I have bumped this to the next release. CMake builds are working fine in my Jenkins CI; I will continue to add this to GitHub actions as part of the 2.5.3 release.

jedwards4b:
I believe that this issue has been addressed. @edwardhartnett do you agree?

jedwards4b:
That is the cmake build no longer needs a netcdf or pnetcdf fortran build.

edwardhartnett:
@jedwards4b the cmake build still required netcdf-fortran:

```
+ cmake -Wno-dev -DNetCDF_C_LIBRARY=/usr/lib/x86_64-linux-gnu/libnetcdf.so -DNetCDF_C_INCLUDE_DIR=/usr/include -DPnetCDF_PATH=/home/runner/pnetcdf -DPIO_ENABLE_FORTRAN=On -DPIO_HDF5_LOGGING=On -DPIO_USE_MALLOC=On -DPIO_ENABLE_LOGGING=On -DPIO_ENABLE_TIMING=Off ..
19
-- The C compiler identification is GNU 7.5.0
20
-- The Fortran compiler identification is GNU 7.5.0
21
-- Detecting C compiler ABI info
22
-- Detecting C compiler ABI info - done
23
-- Check for working C compiler: /usr/bin/mpicc - skipped
24
-- Detecting C compile features
25
-- Detecting C compile features - done
26
-- Detecting Fortran compiler ABI info
27
-- Detecting Fortran compiler ABI info - done
28
-- Check for working Fortran compiler: /usr/bin/mpifort - skipped
29
-- Checking whether /usr/bin/mpifort supports Fortran 90
30
-- Checking whether /usr/bin/mpifort supports Fortran 90 - yes
31
-- Importing CMake_Fortran_utils
32
-- Importing CMake_Fortran_utils - success
33
-- Found MPI_C: /usr/bin/mpicc (found version "3.1") 
34
[...]

PR #1839, opened 2021-01-28
turning on fortran in cmake build

edwardhartnett:
Part of #1728 

Now trying to build FORTRAN with cmake.


edwardhartnett:
I'm dealing with this in another PR.

PR #1857, opened 2021-02-26
fix indicator for pnetcdf support in cmake

### LINKED ISSUE #1856: cmake build issue
If I build with cmake but get the pnetcdf path wrong I see this in the cmake output:
```
# Features
--------
PnetCDF Support:	yes

```
but looking in config.h I see:
```
/* Does PIO support using pnetcdf for I/O? */
/* #undef _PNETCDF */
```

This is confusing and incorrect.

jedwards4b:
Fixes #1856

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -292,7 +292,7 @@ if (PIO_ENABLE_DOC)
   add_subdirectory (doc)
 endif ()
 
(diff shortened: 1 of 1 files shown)


edwardhartnett (approved):
Good catch!

PR #1864, opened 2021-03-08
fix error in check_netcdf call and in detecting dap

### LINKED ISSUE #1861: bug in check_netcdf logic of PIOc_put_vars_tc
The call to check_netcdf here
   https://github.com/NCAR/ParallelIO/blob/master/src/clib/pio_getput_int.c#L1268
is in a section of code only used by iotasks.  If the error handling is set to PIO_BCAST_ERROR
there is a call to MPI_Bcast https://github.com/NCAR/ParallelIO/blob/master/src/clib/pioc_support.c#L687
which cannot be completed because the comp_comm tasks are not present.   

This results in a garbled error message at best.  In the case I am working on I get:
```
MPI_BCAST, Message truncated on receive: An application bug caused the sender to send too much data
```
I suspect this is due to the comp_comm tasks arriving at a different MPI_BCAST call and trying to satisfy the collective.

### LINKED ISSUE #1863: cmake doesn't detect DAP in recent NetCDF versions
I don't know when the change occurred in ${NetCDF_C_INCLUDE_DIR}/includes/netcdf_meta.h but at least as far back as NetCDF-C v4.5.0, 'NC_HAS_DAP' was replaced by 'NC_HAS_DAP2' and 'NC_HAS_DAP4'. The code in cmake/TryNetCDF_DAP.c causes linking with any semi-recent, DAP enabled version of NetCDF to fail because the search for curl and subsequent addition to the NetCDF_C_Libraries list is omitted. I believe line 8 of cmake/TryNetCDF_DAP.c should read:

`#if NC_HAS_DAP==1 || NC_HAS_DAP2==1 || NC_HAS_DAP4==1`

jedwards4b:
Fix issue in detecting dap and in collective comm in getput_int

Fixes #1863 
Fixes #1861

### DIFF
--- a/.github/workflows/autotools.yml
+++ b/.github/workflows/autotools.yml
@@ -27,7 +27,7 @@ jobs:
         sudo apt-spy2 fix --commit
         # after selecting a specific mirror, we need to run 'apt-get update'
         sudo apt-get update
-        sudo apt-get install netcdf-bin libnetcdf-dev doxygen graphviz wget gfortran libjpeg-dev libz-dev openmpi-bin libopenmpi-dev 
+        sudo apt-get install netcdf-bin libnetcdf-dev doxygen graphviz wget gfortran libjpeg-dev libz-dev openmpi-bin libopenmpi-dev
     - name: cache-pnetcdf
       id: cache-pnetcdf
       uses: actions/cache@v2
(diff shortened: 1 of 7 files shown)


jedwards4b:
@edhartnett  Got the actions working again.

PR #1871, opened 2021-05-03
Pass -c to ranlib to include common symbols on macOS

### LINKED ISSUE #1866: Mac OS libpioc.a requires ranlib -c
In building PIO (using `configure`) for testing on MacOS Mojave I have discovered that, in addition to fixes necessary to overcome #1862, the Mac OS linker requires the inclusion of common symbols as definitions with respect to the table of contents. This is achieved by adding the -c flag to the `ranlib` command. The entry from the ranlib man page is as follows:

```
-c     Include common symbols as definitions with respect to the table of contents.  This is seldom the intended behavior
       for linking from a library,  as  it  forces  the  linking  of  a library member just because it uses an
       uninitialized global that is undefined at that point in the linking.  This option is included only because this was
       the original behavior of ranlib.  This option is not the default.
```

One can either edit `libtool` changing RANLIB to "ranlib -c" after `configure` completes or run `ranlib -c src/clib/.libs/libpioc.a` after make completes. Otherwise, `make check` will fail to compile because of undefined references to _err_buffer.

I am using the Intel 19 compilers and Open MPI 2.1.6 on an iMac Pro. On a side note, I also have to disable the `pio_file_fail` general test because it hangs with my configuration.

kgerheiser:
Part of #1866

Intel compiler on macOS doesn't include common symbols in libraries by default. This leads to undefined symbols for `resultlen` and `err_buffer` when building the tests on macOS.


CMakeLists.txt
```diff
@@ -94,6 +94,12 @@ option (PIO_USE_MPISERIAL    "Enable mpi-serial support (instead of MPI)"   OFF)
 option (PIO_USE_PNETCDF_VARD       "Use pnetcdf put_vard "  OFF)
 option (WITH_PNETCDF         "Require the use of PnetCDF"                   ON)
 
+if(APPLE)
```
jedwards4b, line 97: @edhartnett do we need the same for the autoconf version?

kgerheiser: Yes, it looks like it does. In the linked issue he was using `./configure`. I've run into this issue in CMake before, and know the fix. Not sure how to translate it to Autotools.

edwardhartnett: Yes, let's get this merged and then I'll add something to configure.ac whcih I hope will solve the problem.

jedwards4b (approved):

PR #1882, opened 2021-11-17
fixes pnetcdf BYTE read for USE_VARD=OFF

jedwards4b:
Using MPI_BYTE datatype from pnetcdf was broken, this is a fix when USE_VARD=OFF.  I have opened an issue in pnetcdf for USE_VARD=ON.   https://github.com/Parallel-NetCDF/PnetCDF/issues/76

### DIFF
--- a/src/clib/pio_darray_int.c
+++ b/src/clib/pio_darray_int.c
@@ -248,11 +248,12 @@ int get_vard_mpidatatype(io_desc_t *iodesc, MPI_Offset gdim0, PIO_Offset unlimdi
         }
 
 #if PIO_ENABLE_LOGGING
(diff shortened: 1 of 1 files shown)


edwardhartnett:
I had trouble with the byte type in pnetcdf, and most of the tests don't try to test byte on pnetcdf, IIRC. I found what I thought was a bug in pnetcdf, and filed an issue. I don't know if it's been fixed since then...

jedwards4b:
@edwardhartnett I don't see any issue related to the BYTE type in the pnetcdf repo except the one I just opened. Do you know where it is?

jedwards4b:
The response to the issue in the pnetcdf repo is that we should use MPI_SIGNED_CHAR instead of MPI_BYTE.  Testing this now, it may make this PR moot.

jedwards4b:
@edwardhartnett where do I need to look to try turning the BYTE type tests back on for pnetcdf?

jedwards4b:
Closed, https://github.com/NCAR/ParallelIO/pull/1883 is a better fix.

PR #1891, opened 2021-12-03
Refactor subset method

jedwards4b:
Refactors the subset decomposition method to support multiple destinations for each source when reading.  This allows for example halo regions to be populated.    This method so far only works for the subset rearranger and the mapping file cannot be used for write since it would be in-determinant to have multiple sources for the same destination.

### DIFF
--- a/src/clib/pio_darray.c
+++ b/src/clib/pio_darray.c
@@ -161,6 +161,9 @@ PIOc_write_darray_multi(int ncid, const int *varids, int ioid, int nvars,
     pioassert(iodesc->rearranger == PIO_REARR_BOX || iodesc->rearranger == PIO_REARR_SUBSET,
               "unknown rearranger", __FILE__, __LINE__);
 
+    pioassert(iodesc->readonly == 0,"Multiple sources in map for a single destination",__FILE__,__LINE__);
+
+
     /* Check the types of all the vars. They must match the type of
      * the decomposition. */
     for (int v = 0; v < nvars; v++)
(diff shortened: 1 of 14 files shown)


src/clib/pioc.c
```diff
@@ -595,6 +595,10 @@ PIOc_InitDecomp(int iosysid, int pio_type, int ndims, const int *gdimlen, int ma
     /* Remember the maplen. */
     iodesc->maplen = maplen;
 
+    /* check if the decomp is valid for write or is read-only */
```
jedwards4b, line 598: I'm not sure what to do here.   check_compmap gathers the entire compmap to comp task 0 and checks for duplicate values.  If any are found it's an error to use for write and I've added an assert to the darray_write routines to catch this.  But this is a relatively expensive and somewhat memory intense operation so my inclination is to leave it off.

jedwards4b:
@edwardhartnett This is ready to merge, do you want to review?

src/clib/pio_darray.c
```diff
@@ -821,8 +821,9 @@ PIOc_write_darray(int ncid, int varid, int ioid, PIO_Offset arraylen, void *arra
     bufptr = (void *)((char *)wmb->data + arraylen * iodesc->mpitype_size * wmb->num_arrays);
     if (arraylen > 0)
     {
+        PLOG((3, "copying %ld bytes of user data %d", arraylen * iodesc->mpitype_size, iodesc->mpitype_size));
+        printf(( "copying %ld bytes of user data %d num_arrays %d bufptr %x array %x\n", arraylen * iodesc->mpitype_size, iodesc->mpitype_size, wmb->num_arrays, bufptr, array));
```
edwardhartnett: I'm sure you don't want to leave printf statements in...

Is the logging not working for you? Build with --enable-logging and you can add log statements without using printf...

edwardhartnett:
@jedwards4b looks good. I will let you merge.

After this is merged, are you ready for release?

src/clib/pio_darray.c
jedwards4b: I'm confused, this line was already removed.  https://github.com/jedwards4b/ParallelIO/blob/refactor_subset_method/src/clib/pio_darray.c#L825

jedwards4b: logging works but sometimes while debugging I find it easier to enable a single print instead of all of the level 3 prints

jedwards4b:
@edwardhartnett Yes, ready for a release.  Thanks

PR #1906, opened 2022-02-24
update netcdf download path

jedwards4b:
Quick fix for netcdf paths.


.github/workflows/netcdf-4.7.4_hdf5-1.12.0_pnetcdf-12.2_ncint_mpich-3.3_asan.yml
```diff
@@ -43,7 +46,7 @@ jobs:
       uses: actions/cache@v2
       with:
         path: ~/hdf5
-        key: hdf5-${{ runner.os }}-1.12.0-mpich-3.3.2-3
+        key: hdf5-${{ runner.os }}-1.12.0-mpich-${{ env.MPICH_VERSION }}
```
gsjaardema, line 49: Should you have a `env.HDF5_VERSION` here?

jedwards4b: This is a work in progress.  Ultimately I would like to handle these builds in 1 place instead of 10.

PR #1907, opened 2022-02-24
Be explicit about components needed when finding MPI

bartgol:
The reason behind this is subtle. If `COMPONENTS` is not specified, cmake will proceed to look for the corresponding component for each language that has a compiler enabled. Now, that's usually ok, but on GPU machines, this is causing issues, likely related with the fact that we have CUDA flags, but the MPI finding process is not picking up the kokkos compile launcher (which takes care of separating nvcc flags from host compiler flags), so the compiler check fails since CUDA flags are not recognized.

Beside that, I noticed that `find_package (MPISERIAL ...)` is always called specifying the needed components, so this mod also makes MPI-MPISERIAL more symmetrical. And also, it's always a good idea to specify the components you need, if they are known.


bartgol:
Sorry, Github puts PRs in the upstream by default. My apologies.

PR #1929, opened 2022-07-27
add function initdecomp_readonly

jedwards4b:
Originally added this functionality within the initdecomp but this required determining for each
compmap whether it was a readonly compmap.   This was an expensive operation in time and in memory consumption,
so here we have opted for a seperate initdecomp_readonly function so that the user is responsible for indicating if a 
decomp is intended to be readonly.    Added a test in fortran of this functionality and an assert that assures only the SUBSET rearranger is used with this method.


jedwards4b:
@edhartnett @edwardhartnett I'm having trouble satisfying the documentation test - can you please have a look?

PR #1931, opened 2022-08-03
turned off docs in CI

edwardhartnett:
Temporarily turn off documentation build in CI.


jedwards4b:
I figured it out, i don't think that we need this PR.

PR #1980, opened 2024-01-04
Prelim GDAL enabled PIO

msl3v:
Created for diagnosis of shapefile read issue.

MSL

### DIFF
--- a/src/clib/pio_gdal.c
+++ b/src/clib/pio_gdal.c
@@ -1,0 +1,1350 @@
+#include <config.h>
+#include <pio.h>
+#include <pio_internal.h>
+#include <ogr_api.h>
+
+/**
+ * The PIO library maintains its own set of ncids. This is the next
+ * ncid number that will be assigned.
+ */
+extern int pio_next_ncid;
+
+static int
+GDALc_inq_file_metadata(file_desc_t *file, GDALDatasetH hDS, int iotype, int *nvars,
+                  int **rec_var, int **pio_type, int **pio_type_size,
+                  MPI_Datatype **mpi_type, int **mpi_type_size, int **ndims)
+{
+    int mpierr;
+    int ret;
+
+    /* Check inputs. */
+    pioassert(rec_var && pio_type && pio_type_size && mpi_type && mpi_type_size,
+              "pointers must be provided", __FILE__, __LINE__);
+
+    if (hDS == NULL) {
+      return pio_err(file->iosystem, file, -999, __FILE__, __LINE__);
+    }
+    /* How many vars in the file? */
+    if (iotype == PIO_IOTYPE_GDAL)
+    {
+      OGRLayerH hLayer = OGR_DS_GetLayer( hDS, 0 );
+      OGR_L_ResetReading(hLayer);
+
+      OGRFeatureDefnH hFD = OGR_L_GetLayerDefn(hLayer);
+      *nvars = OGR_FD_GetFieldCount(hFD);
+      if (nvars == 0) // empty file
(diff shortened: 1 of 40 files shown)


src/clib/pio_gdal.c
```diff
@@ -0,0 +1,1350 @@
...
+	    break;
+	  case OFTInteger:
+	    (*pio_type)[v] = (int)PIO_INT;
+	    (*pio_type_size)[v] = sizeof(int);
+	    break;
+	  case OFTString:
+	    (*pio_type)[v] = (int)PIO_STRING;
+	    (*pio_type_size)[v] = -1;
+	    break;
+	  // This needs to be done. How do we deal with timestamps etc in GDAL vector fields?
```
jedwards4b, line 83: Can we treat them as separate variables?

src/clib/pio_gdal.c
```diff
@@ -0,0 +1,1350 @@
...
+    if (!name || strlen(name) > NC_MAX_NAME)
+        return pio_err(ios, file, PIO_EINVAL, __FILE__, __LINE__);
+
+    PLOG((1, "GDALc_inq_fieldid fileid = %d name = %s", fileid, name));
+
+    if (ios->async)
+    {
+        if (!ios->ioproc)
+        {
+            int msg = PIO_MSG_INQ_VARID;
```
jedwards4b, line 151: You will need to define unique msg id's for your functions, this is part of the async support.

src/clib/pio_gdal.c
```diff
@@ -0,0 +1,1350 @@
...
+            return check_mpi(NULL, file, mpierr, __FILE__, __LINE__);
+    }
+
+    /* If this is an IO task, then call the netCDF function. */
+    if (ios->ioproc)
+    {
+        switch (file->iotype)
+        {
+        case PIO_IOTYPE_GDAL:
+//            if (ios->io_rank == 0)
```
jedwards4b, line 285: Why is this commented out?  If your IO is limited to a serial interface you only want to open the file on the root io task.

src/clib/pio_gdal.c
```diff
@@ -0,0 +1,1350 @@
...
+
+    /* Get file info based on ncid. */
+    if ((ierr = pio_get_file(fileid, &file)))
+        return pio_err(NULL, NULL, ierr, __FILE__, __LINE__);
+    ios = file->iosystem;
+    if (file->hDS == NULL)
+        return pio_err(NULL, NULL, ierr, __FILE__, __LINE__);
+
+    GDALDatasetH *hDS = file->hDS;
+    OGRLayerH hLayer = OGR_DS_GetLayer( hDS, 0 );
```
jedwards4b, line 426: shouldn't this also be limited to io_root?

src/clib/pio_gdal.c
```diff
@@ -0,0 +1,1350 @@
...
+
+    /* If this task is in the IO component, do the IO. */
+    if (ios->ioproc)
+    {
+        switch (file->iotype)
+        {
+        case PIO_IOTYPE_GDAL:
+            PLOG((2, "Calling GDALCreate io_comm = %d mode = %d fh = %d",
+                  ios->io_comm, mode, file->fh));
+
```
jedwards4b, line 888: I think that these calls should be restricted to io_root.

jedwards4b (changes requested):
It looks like your branch is missing some history and you are trying to bring back a lot of old code - please clean that up and only bring in your intended changes.   It also looks like you are trying to run the interface to gdal on all io tasks when it should only be run on io_root.   When you have made these changes I'll take another look.

jedwards4b:
Superseded by #1981

PR #1981, opened 2024-01-05
Clean import of code for GDAL in ESMF. Compiles/runs in ESMF

msl3v:
Need to diagnose parallelization issues.

### DIFF
--- a/src/clib/pio.h
+++ b/src/clib/pio.h
@@ -19,6 +19,8 @@
 #include <netcdf.h>
 #include <netcdf_meta.h>
 
+#include <gdal.h>
+
 #define NETCDF_VERSION_LE(Maj, Min, Pat) \
     (((NC_VERSION_MAJOR == Maj) && (NC_VERSION_MINOR == Min) && (NC_VERSION_PATCH <= Pat)) || \
      ((NC_VERSION_MAJOR == Maj) && (NC_VERSION_MINOR < Min)) || (NC_VERSION_MAJOR < Maj))
@@ -606,6 +608,13 @@ typedef struct file_desc_t
      * feature. One consequence is that PIO_IOTYPE_NETCDF4C files will
      * not have deflate automatically turned on for each var. */
     int ncint_file;
+
+    /** GDAL specific vars - M.Long */
+    GDALDatasetH *hDS;
+    int dateVarID;     // Index of field with type OFTDate
+    int timeVarID;     // Index of field with type OFTTime
+    int datetimeVarID; // Index of field with type OFTDatetime
+
 } file_desc_t;
 
 /**
@@ -624,7 +633,10 @@ enum PIO_IOTYPE
     PIO_IOTYPE_NETCDF4C = 3,
 
     /** NetCDF4 (HDF5) parallel */
-    PIO_IOTYPE_NETCDF4P = 4
+    PIO_IOTYPE_NETCDF4P = 4,
+
+    /** GDAL (serial only) */
+    PIO_IOTYPE_GDAL = 5
 };
 
 /**
@@ -1356,6 +1368,14 @@ extern "C" {
                              const long long *op);
     int nc_put_vard_ulonglong(int ncid, int varid, int decompid, const size_t recnum,
                               const unsigned long long *op);
+
+    /* These functions are for the GDAL integration layer. MSL - 9/7/2023 */
+    int GDALc_inq_fieldid(int fileid, const char *name, int *varidp);
+    int GDALc_inq_timeid(int fileid, int *timeid); // Is there a field of type OFTDate, OFTTime, or OFTDateTime?
+    int GDALc_openfile(int iosysid, int *fileIDp, GDALDatasetH *hDSp, int *iotype, const char *fname, bool mode);
+    int GDALc_sync(int fileid);
+    int GDALc_shp_get_int_field(int fileid);
+    int GDALc_shp_get_double_field(int fileid, int varid, const size_t *startp, const size_t *countp, double *ip);
 
 #if defined(__cplusplus)
 }
--- a/src/clib/pio_darray.c
+++ b/src/clib/pio_darray.c
@@ -181,7 +181,7 @@ PIOc_write_darray_multi(int ncid, const int *varids, int ioid, int nvars,
 
     /* Run these on all tasks if async is not in use, but only on
      * non-IO tasks if async is in use. */
-    if (!ios->async || !ios->ioproc)
+    if ((!ios->async || !ios->ioproc) && (file->iotype != PIO_IOTYPE_GDAL))
     {
         /* Get the number of dims for this var. */
         PLOG((3, "about to call PIOc_inq_varndims varids[0] = %d", varids[0]));
@@ -947,6 +947,10 @@ PIOc_read_darray(int ncid, int varid, int ioid, PIO_Offset arraylen,
     case PIO_IOTYPE_PNETCDF:
     case PIO_IOTYPE_NETCDF4P:
         if ((ierr = pio_read_darray_nc(file, iodesc, varid, iobuf)))
+            return pio_err(ios, file, ierr, __FILE__, __LINE__);
+        break;
+    case PIO_IOTYPE_GDAL:
+        if ((ierr = pio_read_darray_shp(file, iodesc, varid, iobuf)))
             return pio_err(ios, file, ierr, __FILE__, __LINE__);
         break;
     default:
--- a/src/clib/pio_file.c
+++ b/src/clib/pio_file.c
@@ -283,6 +283,10 @@ PIOc_closefile(int ncid)
             ierr = ncmpi_close(file->fh);
             break;
 #endif
+	case PIO_IOTYPE_GDAL:
+	  ierr = GDALClose((void*)file->hDS);
+	  printf("GDALClose ierr: %d\n",ierr);
+	  break;
         default:
             return pio_err(ios, file, PIO_EBADIOTYPE, __FILE__, __LINE__);
         }
--- a/src/clib/pio_gdal.c
+++ b/src/clib/pio_gdal.c
@@ -1,0 +1,791 @@
+#include <config.h>
+#include <pio.h>
+#include <pio_internal.h>
+#include <ogr_api.h>
+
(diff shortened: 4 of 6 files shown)


src/clib/pio.h
```diff
@@ -19,6 +19,8 @@
 #include <netcdf.h>
 #include <netcdf_meta.h>
```
jedwards4b, line 21: Should also include an #ifdef here.

src/clib/pio_file.c
```diff
@@ -283,6 +283,10 @@ PIOc_closefile(int ncid)
             ierr = ncmpi_close(file->fh);
             break;
 #endif
+	case PIO_IOTYPE_GDAL:
```
jedwards4b, line 286: I think that you need a line like 274 here:
if (ios->io_rank == 0)
this should only be called on the root io task.

src/clib/pio_darray.c
```diff
@@ -949,6 +949,10 @@ PIOc_read_darray(int ncid, int varid, int ioid, PIO_Offset arraylen,
         if ((ierr = pio_read_darray_nc(file, iodesc, varid, iobuf)))
             return pio_err(ios, file, ierr, __FILE__, __LINE__);
         break;
+    case PIO_IOTYPE_GDAL:
```
jedwards4b, line 952: Will need #ifdef

src/clib/pio.h
```diff
@@ -606,6 +608,13 @@ typedef struct file_desc_t
      * feature. One consequence is that PIO_IOTYPE_NETCDF4C files will
      * not have deflate automatically turned on for each var. */
     int ncint_file;
+
```
jedwards4b, line 611: #ifdef here too

src/clib/pio_gdal.c
```diff
@@ -0,0 +1,791 @@
...
+
+    /* Fill in some file values. */
+    file->fh = -1;
+    file->iotype = *iotype;
+    file->iosystem = ios;
+    file->writable = (mode & PIO_WRITE) ? 1 : 0;
+
+    /* Set to true if this task should participate in IO (only true
+     * for one task with netcdf serial files. */
+    if (file->iotype == PIO_IOTYPE_GDAL ||
```
jedwards4b, line 254: shouldn't this be && not ||?

src/clib/pio_gdal.c
```diff
@@ -0,0 +1,791 @@
...
+            return check_mpi(NULL, file, mpierr, __FILE__, __LINE__);
+    }
+
+    /* If this is an IO task, then call the netCDF function. */
+    if (ios->ioproc)
+    {
+        switch (file->iotype)
+        {
+        case PIO_IOTYPE_GDAL:
+//            if (ios->io_rank == 0)
```
jedwards4b, line 296: You are writing a serial interface - you should only call your GDAL functions if io_rank == 0

jedwards4b (changes requested):
You are writing a serial interface - you should only call GDAL from io_root
emulate the netcdf serial interface.  NetCDF files are read on io_root and then results are distributed to all tasks.   By the way - what does your test look like?  You will want to add it to the tests/cunit directory.

jedwards4b:
Also you don't need to create a new PR with additional changes, just push them back to the same branch.   How is the GDAL library available?  Can it be obtained with apt-git?  Just trying to figure out how to add it to the github testing workflow.

src/clib/pio_darray.c
msl3v: Hey. Yeah. I tried to set up an #ifdef. It's on the list of to-do

msl3v:
GDAL is available in several forms. It is available in my Fedora repo via dnf as an RPM. The source is also available, and is not complex to compile. We have it on the NCCS machines here at NASA.

src/clib/pio_gdal.c
msl3v: Yep. Thanks.

msl3v:
@jedwards4b, the ESMF test results show that for a file with 4 points, all 4 points are being read correctly on io_root. The data being read are [1 2 3 4]. If I run on 4 procs, the local buffers are [1] [1] [1] [1]. If I run on 2 procs, the local buffers are [1 2] [1 2]. So it looks like MPI is sending/recving but that indexing isn't correct.

jedwards4b:
If you could provide your test program I'll give it a try.   Please remove reference to ESMF, let's take that library out of the equation for now.

msl3v:
ESMF does all the calling to PIO, and it would be a complex task (for me) to uncouple. ESMF starts in Fortran, can then calls into C bindings, sets up arrays etc, and then calls PIO. I'll see if I can isolate the PIO-specifc pieces but we may have to iterate.

msl3v:
Is there an existing test in PIO that I could modify for this?

msl3v:
I've added test_gdal.c in the most recent commit to github.com/msl3v/ParallelIO:msl3v/GDAL_clean
It segfaults in pio_swapm() and I'm not able to diagnose why. If you have any insight it would be appreciated. Thank you.

jedwards4b:
Great thanks - I have a day full of meetings today, but I'll try to get to this as soon as I can.

jedwards4b:
What version of gdal is required - my system has gdal/3.7.1 built with gcc - can I test with that?

msl3v:
3.7.1 works. Here's my cmake cmd

env CC=mpicc FC=mpifort CFLAGS="-g -fPIC" cmake -DCMAKE_INSTALL_PREFIX=../Install -Wno-dev -DPIO_ENABLE_TIMING=OFF -DPIO_ENABLE_EXAMPLES=OFF -DPIO_ENABLE_FORTRAN=OFF -DPIO_ENABLE_LOGGING=ON -DCMAKE_BUILD_TYPE=DEBUG -DNetCDF_C_INCLUDE_DIR=/usr/lib64/gfortran/modules -DNetCDF_C_LIBRARY=/usr/lib64/libnetcdf.so -DWITH_PNETCDF=off -DGDAL_INCLUDE_DIR=<path_to_gdal>/include -DGDAL_LIBRARY=<path_to_gdal_lib64>/libgdal.so ../ParallelIO

jedwards4b:
The first problem is because your test doesn't run from the source directory and so it
can't find file data/cb_2018_us_region_20m.shp in the relative path.   i'll look into how to do this in the build, but for now you can create a link from tests/cunit/data to build/tests/cunit/data.

msl3v:
Interesting. It opens, queries and reads data from the file in my local version. Curious what's different

msl3v:
Is it maybe that I'm running the test directly from tests/cunit?
I'll create the link and push the change.

jedwards4b:
Looks like you found it - that's great!

msl3v:
OK. The issue with the segfault was that the recvbuf was NULL. It was fixed with a simple line in test_gdal.c
I'll keep banging on it to see what the recvbuf data look like.

msl3v:
@jedwards4b 
In my pio_log output, for 4 data points on 4 MPI tasks, I get the following output:

0 sendcounts[0] = 1 sdispls[0] = 0 sendtypes[0] = 27742288 recvcounts[0] = 1 rdispls[0] = 0 recvtypes[0] = 27745072
0 sendcounts[1] = 1 sdispls[1] = 0 sendtypes[1] = 27743024 recvcounts[1] = 0 rdispls[1] = 0 recvtypes[1] = 2085777792
0 sendcounts[2] = 1 sdispls[2] = 0 sendtypes[2] = 27743696 recvcounts[2] = 0 rdispls[2] = 0 recvtypes[2] = 2085777792
0 sendcounts[3] = 1 sdispls[3] = 0 sendtypes[3] = 27744368 recvcounts[3] = 0 rdispls[3] = 0 recvtypes[3] = 2085777792

Is this telling me that only one data point is being received (line 1)?

msl3v:
Ah! Something to do with iodesc->ndof

msl3v:
@jedwards4b I'm at a loss. I can keep trying but it's just not clear what the problem might be. Is it the case that the data from the serially-read vector are broadcast via the MPI_AlltoAllw() call in rearrange_io2comp()->pio_swapm()?

jedwards4b:
Yes that's correct, I can try to have another look at this tomorrow.

msl3v:
Is it the case, then, that the first line in the output above
   0 sendcounts[0] = 1 sdispls[0] = 0 sendtypes[0] = 27742288 recvcounts[0] = 1 rdispls[0] = 0 recvtypes[0] = 27745072
is saying that only task 0 is set to receive any data, since all the rest of the receive params are zero?

jedwards4b:
yes, I will try to look at this again after lunch.

msl3v:
No hurry. I'm just banging away at the code & posting what comes up.

msl3v:
@jedwards4b just an update: Still no progress. It looks like the iodesc->map might not be getting set up correctly. MPI doesn't know to send what element to which proc. ... maybe. But I still haven't interpreted what is actually happening here. I'll keep trying.

jedwards4b:
I think that you should walk through the netcdf serial test step by step - it should be exactly the same as this one.

msl3v:
@jedwards4b 
It's not clear which test is the mpi serial test. I've modeled the GDAL test off test_darray.c 
Is the serial test test_simple?

Also, in a totally vanilla version of ParallelIO, all *darray* and the test_simple tests fail on my laptop. So maybe that is itself part of the problem. :-/

jedwards4b:
That does seem like a problem.   On my system all of the tests are passing except the gdal test.  In the test_darray.c there is a loop over num_flavors at line 110.  One of those flavors is netcdf-serial (PIO_IOTYPE_NETCDF).   Your test should behave exactly the same as test_darray.c does with this flavor.

msl3v:
Switched to another system. All tests pass (from remotes/origin/main). It seems to be a local MPI problem on my laptop.
Thanks for the pointer re test_darray.c. Onward.

msl3v:
Yeah. It may simply be that my laptop openMPI was upgraded (to v 4.1.9) and no longer works with ParallelIO. Switched to MPICH and things seem to be working, including the GDAL read :-/

msl3v:
Closing this. Will recreate with a parallel-enabled branch.

PR #2002, opened 2025-02-14
Fix issues that prevent compilation in Cygwin

harmenwierenga:
[See my PR for ESMF](https://github.com/esmf-org/esmf/pull/348). ESMF uses ParallelIO, and it fails to build through cygwin because of a few small issues in ParallelIO. Here, I try to address those:

- In the src/clib/CMakeLists.txt, there is a check that long long and size_t are the same size. On cygwin, these sizes cannot be retrieved, which makes them both empty strings. This causes the check to fail, because empty strings cannot be parsed as numbers so ${SIZEOF_SIZE_T} EQUAL ${SIZEOF_LONG_LONG} evaluates to false in CMake. I modified the check to simply test if they are the same, not if they are parsable numbers and the same.
- In the src/clib/pioc_support.c there is an include of execinfo.h. This header is missing on many platforms, including Cygwin. I have added a more portable check to see if the header is available, and I have excluded the include and the use of the backtrace functionality from that file when execinfo.h is not available. The backtrace functionality appears to be added as some useful logging for debugging purposes, but not as a necessary feature of the library, so I hope I was correct that it could simply be turned off. I have tried to suppress any warnings of unused variables.

Thank you for considering this, and please let me know if I should make changes!

### DIFF
--- a/src/clib/CMakeLists.txt
+++ b/src/clib/CMakeLists.txt
@@ -175,8 +175,8 @@ endif ()
 
 include(CheckTypeSize)
 check_type_size("size_t" SIZEOF_SIZE_T)
-CHECK_TYPE_SIZE("long long" SIZEOF_LONG_LONG)
-if (NOT ${SIZEOF_SIZE_T} EQUAL ${SIZEOF_LONG_LONG})
+check_type_size("long long" SIZEOF_LONG_LONG)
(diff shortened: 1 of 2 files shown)


jedwards4b (approved):
Thanks for your contribution!

PR #2023, opened 2026-03-13
Include lib/CMAKE_LIBRARY_ARCHITECTURE in LibFind

danrosen25:
## Issue
apt installs NetCDF libraries to `/usr/lib/<architecture>` instead of `/usr/lib`. It's possible to configure Spack to use an externally installed NetCDF library using `packages.yaml`. So if you want parallelio to find libraries installed by apt that are externally configured for Spack then you need to search the `/usr/lib/<architecture>` directory. Note that I am using apt to install libraries because I'm building container images using emulation for multiple architectures. If I were to build each library (instead of using binaries provided by apt) then it would take 12 hours (an estimate) and that's not possible for automated deployment from GitHub runners.

example packages.yaml file
```yaml
packages:
  netcdf-c:
    externals:
    - spec: netcdf-c@4.9.3%gcc@15.2.0
      prefix: /usr
```

## Changes
This fix adds lib/${CMAKE_LIBRARY_ARCHITECTURE} to the search suffixes in LibFind.cmake. It also fixes an issue setting ${PKGCOMP}_FOUND in the same file.

## Testing
I've lightly tested this in a dev container. It can use more robust testing.

### DIFF
--- a/cmake/LibFind.cmake
+++ b/cmake/LibFind.cmake
@@ -254,7 +254,7 @@ function (find_package_component PKG)
                 find_library (${PKGCOMP}_LIBRARY
                               NAMES ${${PKGCOMP}_LIBRARY_NAMES}
                               PATHS ${${PKGCOMP}_PREFIX}
-                              PATH_SUFFIXES lib
+                              PATH_SUFFIXES lib lib/${CMAKE_LIBRARY_ARCHITECTURE}
                               NO_DEFAULT_PATH)
 
                 # If found, check if library is static or dynamic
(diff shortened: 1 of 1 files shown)


danrosen25:
FYI, the netcdf_hdf5_no_pnetcdf_ncint_mpich action failed during apt-get install; it failed fetching libcurl4-gnutls-dev. The recommended fix is to run apt-get update first to update the package list.

jedwards4b (approved):

jedwards4b:
I think that in spack we need to indicate that pio versions older than 2.6.5 are not compatible with netcdf versions newer than 4.9.2

danrosen25:
@jedwards4b I can add a conflict message to the spack recipe
Is this sufficient?
`conflicts("^netcdf-c@4.9.3:", when="@:2.6.4", msg="netcdf-c@4.9.3+ is incompatible with parallelio versions prior to 2.6.5")`

jedwards4b:
That looks great, thank you

jedwards4b:
I've created pio2_6_10 with this change.  Thanks

danrosen25:
@jedwards4b 
Thanks, the conflict works as expected but it's a rabbit hole. It forces building c-blosc-1.21.6 and that fails on my system, probably a compiler error with gcc-15.

==> Warning: Skipping build of netcdf-c-4.9.2-faklxt3nctalpcpp5ejwsjiwyguerutk since c-blosc-1.21.6-4fypsbdske4f7fseu4lfjxjbj54zihyt failed
==> Warning: Skipping build of parallelio-2.6.4-f7wlcj3lqeyhyr3xy3pyyroz7eq4e3q5 since netcdf-c-4.9.2-faklxt3nctalpcpp5ejwsjiwyguerutk failed

I've put in the PR to spack-packages but did you want me to bump the version or do you want to wait?
https://github.com/spack/spack-packages/pull/3840

jedwards4b:
I did bump the pio tag to pio2_6_10.

PR #2024, opened 2026-03-13
Update apt package index before installation

danrosen25:
This should be the fix for the failed workflow and others


jedwards4b (approved):
thanks

